PR #8, opened 2025-11-06
added telegram support

syswraith:
how it works
- hit the https://t.me/username endpoint
- check for the `body > div.tgme_page_wrap > div.tgme_body_wrap > div > div.tgme_page_extra` selector
- if it exists, the account exists


kaifcodec (approved):
Nice work!

kaifcodec:
Nice work, but always try to keep the `if __name__ == "__main__":` block for making it easier to local testings.
Also feel free to check out these issues #12 #13  and if you are familiar with them a PR is always welcome!

Issue #12, opened 2025-11-07
Add epic games to gaming 


Meletion:

kaifcodec:
@thecodealchemy Yeah you can take this (epic games).

PR #20, opened 2025-11-12
added discord under social 

Jmr13:
1. Discord has a limit of anonymous API request, so it'll not work after a while
2. If username exists, the "taken" key from the response is True, otherwise False

=== Sample cURL request ===
curl -X POST 'https://discord.com/api/v9/unique-username/username-attempt-unauthed' -H 'Content-Type:application/json' -d '{
	"username": "SampleUsername"
}'

### DIFF
--- a/user_scanner/social/discord.py
+++ b/user_scanner/social/discord.py
@@ -1,0 +1,44 @@
+import httpx
+from httpx import ConnectError, TimeoutException
+
+def validate_discord(user):
+    url = "https://discord.com/api/v9/unique-username/username-attempt-unauthed"
+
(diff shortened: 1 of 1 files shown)


kaifcodec:
@Jmr13  Thank you for this perfect PR and really appreciate your work for finding out the private API of discord that is really there for this specific purpose. So it's less likely to be error prone.

If you’re up for it, feel free to take a look at issues #12 and #13 next, they’d both benefit from your touch.

Jmr13:
Glad to help! Will check other issues thanks 😁

PR #25, opened 2025-11-16
Fix header compression & Roblox case sensitivity

VamatoHD:
This PR addresses two issues:

1. Responses from Roblox are case-sensitive. With this PR, they are handled as case-insensitive.
2. Due to the response being compressed, Chess.com and Monkeytype were generating an error. This was due to the header allowing compressed encoding; now they are sent without compression.


kaifcodec:
@VamatoHD Thanks for the PR! The Roblox changes are really helpful and necessary, especially the case-insensitive check. The other updates weren’t strictly required, but using "identity" for Accept-Encoding is a safer approach and a good precaution.

If you’d like, you can also explore other modules, there’s room for improvements across the project.

VamatoHD:
I changed the header into "identity" as the responses were compressed (at least on my machine).

I'll look into improving the project!

kaifcodec:
Thanks for checking that! If the API was returning compressed responses on your setup, then switching to "identity" definitely makes sense. It keeps things predictable across different environments.

And sure, feel free to explore the project anytime. There’s plenty of space to improve things across modules.

Issue #27, opened 2025-11-17
Why not use threading?


VamatoHD:
Hi @kaifcodec,

I noticed the code checks each website one by one, which will get considerably slower as more checks are added. Is there a reason for not using threading/async?

Thanks!

kaifcodec:
@VamatoHD Thanks for bringing this up!

I had actually been thinking about implementing asyncio and concurrency for each category a few days ago. The idea was to improve performance as the number of modules grows with incoming PRs, and also to make the output feel more dynamic, almost like a live stream. I’ve been tied up with other projects, so I haven’t had a chance to work on it yet. If you’re interested, this could be a great area to contribute, I’d be happy to collaborate or review any changes.

VamatoHD:
@kaifcodec, I am genuinely interested in implementing it; however, I am also preoccupied with other projects. I'll for sure implement it when I have some free time!

PR #28, opened 2025-11-19
Implement threading

VamatoHD:
Implemented threading (#27)

- The output order is preserved even if threads complete at different times.
- **Only tested briefly** — Automatic tests could be added in the future.
- There may be more efficient solutions, but this version is already an improvement.


kaifcodec:
@VamatoHD Thank you for the PR and It was really needed, the changes look clean and working.
Maybe I will tweak it a little bit later but right now it's okay to be merged.

PR #33, opened 2025-11-28
Added Steam and Itch.io

VamatoHD:
Implemented username checking for:
1. Steam*
2. Itch.io

*Since steam's API requires an API key, I opted for a simple web scrap. There may be better ways to validate.

Also, just as a thought, most validators just do a simple “status” check. Maybe merging them into a function would be a better and cleaner option.

### DIFF
--- a/user_scanner/core/orchestrator.py
+++ b/user_scanner/core/orchestrator.py
@@ -23,7 +23,7 @@ def worker_single(module, username, i):
 
     func = next((getattr(module, f) for f in dir(module)
                  if f.startswith("validate_") and callable(getattr(module, f))), None)
-    site_name = module.__name__.split('.')[-1].capitalize()
+    site_name = module.__name__.split('.')[-1].capitalize().replace("_",".")
     if site_name == "X":
         site_name = "X (Twitter)"
 
--- a/user_scanner/creator/itch_io.py
+++ b/user_scanner/creator/itch_io.py
@@ -1,13 +1,13 @@
 import httpx
 from httpx import ConnectError, TimeoutException
 
-def validate_steam(user):
+def validate_itch_io(user):
     """
-    Checks if a steam username is available.
+    Checks if a itch.io username is available.
     Returns: 1 -> available, 0 -> taken, 2 -> error
     """
 
-    url = f"https://steamcommunity.com/id/{user}/"
+    url = f"https://{user}.itch.io"
 
     headers = {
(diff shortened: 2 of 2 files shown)


kaifcodec:
@VamatoHD Does steam fit in `gaming/` category?

As far as I got the details by searching about it, It looks like a digital distribution platform for PC games. It also looks like a service that allows users to buy, download, play, and manage video games, and also includes community features, cloud saves, and other tools. 

Anyways I also want to discuss about what you suggested like, to merge them in a function.
Can you explain what advantages are you seeing if we implement that?

VamatoHD:
I'd say Steam belongs in the `gaming/` category since its core features are around games, such as the game library, updates, achievements, etc. Even though it has community features like chats, user forums, they are secondary tools used to enhance the gaming experience.

In the topic of merging into a function, having a generic function for requests or for status checking in the `orchestrator.py` wouldn't make a big difference in terms of performance. It would be primarily to follow a principle called DRY (Don't repeat yourself).

kaifcodec:
@VamatoHD, I've reviewed the code and it's ready to merge.
Regarding your DRY suggestion for `core/orchestrator.py`: I appreciate the feedback, but I need clarity on the scope. We already have `run_module_single()` and `run_checks_category()`.

Are you suggesting a helper function for 

1. Wrapping the **request execution/error handling** (the `try/except` around `validate_`)?
2. Just the **status result formatting** (mapping `1`, `0`, or `else` to the color-coded output)?

Clarifying the specific duplicated logic would help me implement the solution.

Feel free to open the Pull Request whenever you like, and we can continue discussing this refactoring suggestion in the issue tracker.

VamatoHD:
Maybe my English was misleading. My thought was that since most `validate_` functions just fetch from a URL and do a status check, they could be compressed into a single function. Perhaps a generic function like `status_check(url, available_status, unavailable_status, header="")`, where `available_status` and `unavailable_status` are lambda functions that compare the response status and the header is optional, does the trick.

kaifcodec:
@VamatoHD Got it, anyways if you are willing to open a PR with those modifications then it would be better, either way I will look into that matter.

Anyways I was thinking about many more options and features here like a `--osint` mode it will use the same modules just shows with different visual letters like instead of "Available" or "Taken" it would go like "Target found" or something else and later maybe email id OSINT.

More like `holehe` and `sherlock` in one tool with many other features and with very low dependencies.

PR #46, opened 2025-12-01
Add Facebook username validation support (social category)

ali-m07:
This PR adds Facebook username validation to the social category.

- Implements \ using the shared \ helper
- Uses the pattern https://www.facebook.com/<username> for profile URLs
- Treats HTTP 404 as available and HTTP 200 as taken

### DIFF
--- a/user_scanner/social/facebook.py
+++ b/user_scanner/social/facebook.py
@@ -1,0 +1,27 @@
+from ..core.orchestrator import status_validate
+
+
+def validate_facebook(user: str) -> int:
+    """
+    Checks if a Facebook username is available.
+
+    Strategy:
+    - Facebook profile URLs follow the pattern: https://www.facebook.com/<username>
+    - Existing usernames typically return HTTP 200.
+    - Non-existing usernames are expected to return HTTP 404 or a similar not-found status.
+    - We rely on HTTP status codes using status_validate.
+    """
+    url = f"https://www.facebook.com/{user}"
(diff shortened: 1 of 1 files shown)


kaifcodec:
@ali-m07 Thanks for the PR, great job, it works fine!

One suggestion: Facebook enforces strict username rules (allowed symbols, length, etc.). To make this validator more accurate, could you add a regex-based pre-check before calling status_validate()?

Suggested flow:

- If the username matches the Facebook pattern → call status_validate()
- If it doesn’t → return 2 with a clear message explaining what constraint failed

This avoids unnecessary requests and gives the user better feedback.
If you run into any issues with showing the validation error message, feel free to ask, and once updated, you can push the changes here and reopen the PR.

SylvainM98:
Hey @ali-m07, I ran into this while working on the project.

I tested the `facebook.com/{username}` approach and it no longer works — Facebook now returns HTTP 400 for all direct profile URL requests (existing, non-existing, invalid) regardless of headers, cookies, or user agent. This means `status_validate(url, 404, 200)` always returns an error.

I opened #263 with an alternative approach that uses the account recovery lookup endpoint (same one used by the existing email scanner module). It reliably differentiates existing from non-existing usernames and includes the regex pre-validation that @kaifcodec requested.

kaifcodec:
Closing this PR as Facebook username validation has been added correctly in #323

Issue #45, opened 2025-12-01
Add username support for liberpay.com


kaifcodec:
- Add username validation for liberpay.com
- Add it to `donation/` category
- Possible url `https://en.liberapay.com/{username}`
- Most likely a simple 404/200 status check

ganeshnikhil:
Implimented Liberapay Validation Function in donation category.

camtisocial:
how'd you get around cloudflare being a pain? @ganeshnikhil

kaifcodec:
@ganeshnikhil Your PR about `liberpay` validation have issues check out the PR conversation #50 

But as @camtisocial pointed out that it have cloudflare verification then normal status validation won't work in that case we might look for the internal private API that is called during registration of a new account.

json-hunter07:
For anyone interested in working on this issue, a quick heads-up:

Liberapay returns a `403` when using the default headers from `core/orchestrator.py` 

Recommended approach:
- Capture the actual headers from a real Liberapay get request , OR test locally to confirm whether the response is compressed (recommended).
- If it’s compressed, update `Accept-Encoding` to only `gzip`.
- Use `generic_validate()` since this requires custom headers (not `status_validate()` you can take hints from `gaming/monkeytype.py` about how to use `generic_validate()` function for custom headers).

If you follow the above steps, the validation works fine locally. Whoever picks this issue, feel free to open a PR after verifying it locally first.

@VamatoHD If you are interested this wouldn't take you long.

VamatoHD:
Alright, I'll take this.
If this ends up like #31, where it aggressively blocks non-browser requests, it will have to wait. Otherwise I'll do it as soon as possible, along with the type hinting for functions.

kaifcodec:
@VamatoHD As @json-hunter07 mentioned I think just a simple change in the `Accept-Encoding` like keeping the value only "gzip" most likely will work and the headers might be a little different.

Maybe this header below could work,

```bash
headers = {
  'User-Agent': "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Mobile Safari/537.36",
  'Accept': "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7",
  'Accept-Encoding': "gzip",
[...]

PR #50, opened 2025-12-02
Liberapay Validation Function

ganeshnikhil:
added new function to validate liberpay

### DIFF
--- a/user_scanner/gaming/minecraft.py
+++ b/user_scanner/gaming/minecraft.py
@@ -1,0 +1,24 @@
+from ..core.orchestrator import status_validate
+
+
+def validate_minecraft(user):
+    """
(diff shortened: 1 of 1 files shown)


kaifcodec:
@ganeshnikhil  Thank you for this PR.

But I think you misunderstood the concept mentioned in the issue #45 .
You were supposed to create a different file named `liberpay.py` and add it to `.donation/` directory.

I would suggest you to go through the contributing guidelines and re open this pull request.

kaifcodec:
@ganeshnikhil Thank you for the updates but Its recommended to open a new PR with only the addition of the liberpay.py 
If this PR is merged then there will be total of 3 commits in history with all the hit and trials.

PR #51, opened 2025-12-02
updated validate_facebook with check and specific error message

ganeshnikhil:
Validates a Facebook username using a regex pre-check and then an external status check.


kaifcodec:
@ganeshnikhil Thanks for this PR and it looks clean and the regex implementation looks correct but I need some time to review this and test it locally.

Till then you can check around the codes for further improvements.

PR #52, opened 2025-12-02
added username availability check for minecraft

camtisocial:
added check for minecraft using status_validate


kaifcodec:
@camtisocial Thank you for the PR and minecraft is really a good addition in gaming category.

- I saw that you also added status code `302` equivalent to `200` does minecraft really returns 302 for taken usernames?

camtisocial:
@kaifcodec nah, yeah that was unnecessary, I read the mojang docs and removed the 302/switched to a more reliable endpoint while i was at it

PR #54, opened 2025-12-02
Add Liberapay function

ganeshnikhil:


json-hunter07:
@ganeshnikhil I think you didn’t test this locally before opening the PR, the request keeps failing with status `403` because the headers don’t match what the Liberpay server expects.

The default headers in `core/orchestrator.py` won’t work here. We need to capture the actual headers used during a real Liberpay API call or at least run tests locally to confirm whether the server returns a compressed response.  
If it does, we should update `Accept-Encoding` to only `gzip`.

Also, this case must use the `generic_validate()` function so we can apply the custom headers properly.

If you understand the changes needed, you can go ahead and open a new PR with the fix. Just make sure to test it locally first.  
Otherwise, feel free to wait and we’ll update it later on our end.

PR #56, opened 2025-12-02
Feature/facebook

ganeshnikhil:
added facebook with validator

### DIFF
--- a/user_scanner/social/facebook.py
+++ b/user_scanner/social/facebook.py
@@ -1,0 +1,52 @@
+
+from ..core.orchestrator import status_validate
+import re 
+
+def validate_facebook(user:str) -> int:
+    """
+    Validates a Facebook username using a regex pre-check and then an external status check.
+    """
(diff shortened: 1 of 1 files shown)


kaifcodec:
@ganeshnikhil I appreciate your constant effort towards `liberapay` and `facebook` but while testing it locally I found some issues in both of your commits.

- `liberapay` is returning `302` status code for existing usernames
- You have used `status_validate()` for facebook assuming it to return `404` for non existing usernames but facebook returns `200` for both existing and non existing usernames, so you have to inspect the response text that it returns after `GET` request and look for some unique keywords that will confirm if the username is available to take or not.
- In this case you have to use `generic_validate()` function.
- Anyways your regex validation is perfectly done.

Most importantly the `liberapay` issue #45 has been closed by #57 so you don't have to implement `liberapay` next time if you want to fix the `facebook` issue. Just re-fork and open a fresh new PR.

Issue #60, opened 2025-12-03
Meaning of Error


ankitom:
What does "Error" mean ?

> == Dev SITES ==
>   [✔] Codeberg: Available
>   [✔] Cratesio: Available
>   [✔] Dockerhub: Available
>   [✘] Github: Taken
>   [✔] Gitlab: Available
>   [✔] Launchpad: Available
>   [!] Npmjs: Error
>   [✔] Replit: Available
> 
> == Social SITES ==
>   [✔] Bluesky: Available
>   [✔] Discord: Available
>   [✘] Instagram: Taken
>   [✔] Mastodon: Available
>   [✘] Pinterest: Taken
>   [!] Reddit: Error
>   [✘] Snapchat: Taken
>   [✘] Telegram: Taken
>   [✘] Threads: Taken
>   [!] X (Twitter): Error
>   [✘] Youtube: Taken
> 
> == Creator SITES ==
>   [✔] Devto: Available
>   [✔] Hashnode: Available
>   [✔] Itch.io: Available
>   [✔] Kaggle: Available
>   [!] Medium: Error
>   [✔] Patreon: Available
>   [!] Producthunt: Error
>

kaifcodec:
@ankitom The error here means either there is a timeout while making the request to those sites, they are down or the username that you used could contain some characters that are not supported by those sites which returns status code other than `404` or `200`.

We have custom error message for some of the sites but right now not implemented to all the sites, so it fails silently.

Issue #31, opened 2025-11-25
Add username validation for Ko-fi


kaifcodec:
- Add username validation for **Ko-fi** in a new category `donation/`.
- Update `__main__.py` and `core/orchestrator.py` to recognize the new category.
- Ensure the new module follows the same structure and conventions as other social/creator modules.

VamatoHD:
@kaifcodec After some brief testing, I found out **Ko-fi aggressively blocks non-browser requests**. The only solution is to use a library like **Playwright** to mimic a browser and check if the user exists. Since this library is kind of heavy, it would need to be optional, and implementing it is hard (the part of just opening one browser for every site that requires it) but doable.

kaifcodec:
@VamatoHD If ko-fi blocks the non-browser requests then I think it would better not to implement it in our list, atleast not now. 

You are correct  that selenium or playwright like headless browsers are way too heavy to use just for simple username checks.

PR #63, opened 2025-12-03
PoC - Replace integer status with result enum

VamatoHD:
This is a **proof-of-concept** implementation showing how the current status integer could be replaced with a Result enum.

**Goals of this PoC:**
1. **Improve readability**: Instead of a simple `1`, we would write `Result.available()`;
2. **Results store the error message**: This way the error could be cleaner for the user since a message like “Too many requests” would be displayed instead of just “error”;
2.1. This **fixes the concurrency issues** that would happen if a print was done inside the `validate_<site>`, since the results would be printed in order;
3. Could enable the use of the unused **verbose flag**;

I'm still not sure about the direction to take, so I just made a boilerplate of the Result enum and a demonstration of how it could be applied in `x.py`. 

**Collaborators are welcome to modify, restructure, and rewrite it as needed.**

### DIFF
--- a/user_scanner/core/result.py
+++ b/user_scanner/core/result.py
@@ -37,7 +37,7 @@ class Result:
         return cls(Status.ERROR, reason)
 
     @classmethod
-    def from_number(cls, i: int, reason: str | None):
+    def from_number(cls, i: int, reason: str | Exception | None = None):
         try:
             status = Status(i)
         except ValueError:
@@ -59,6 +59,9 @@ class Result:
         #Format the exception
         msg = humanize_exception(self.reason)
         return f"{type(self.reason).__name__}: {msg.capitalize()}"
+
+    def __str__(self):
+        return self.get_reason()
 
     def __eq__(self, other):
         if isinstance(other, Status):
--- a/user_scanner/creator/hashnode.py
+++ b/user_scanner/creator/hashnode.py
@@ -1,5 +1,4 @@
 import httpx
-import json
 from user_scanner.core.result import Result
 
 
--- a/user_scanner/creator/itch_io.py
+++ b/user_scanner/creator/itch_io.py
@@ -2,11 +2,6 @@ from user_scanner.core.orchestrator import status_validate
 
 
 def validate_itch_io(user):
-    """
-    Checks if a itch.io username is available.
-    Returns: 1 -> available, 0 -> taken, 2 -> error
-    """
-
     url = f"https://{user}.itch.io"
 
     return status_validate(url, 404, 200, follow_redirects=True)
--- a/user_scanner/creator/medium.py
+++ b/user_scanner/creator/medium.py
@@ -1,5 +1,5 @@
-import httpx
-from httpx import ConnectError, TimeoutException
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
 
 
 def validate_medium(user):
@@ -10,24 +10,19 @@ def validate_medium(user):
         'Accept': "text/html",
     }
 
-    try:
-        response = httpx.get(url, headers=headers, timeout=3.0)
-
+    def process(response):
         if response.status_code == 200:
             html_text = response.text
 
             username_tag = f'property="profile:username" content="{user}"'
 
             if username_tag in html_text:
-                return 0
+                return Result.taken()
             else:
-                return 1
-        return 2
+                return Result.available()
+        return Result.error()
 
-    except (ConnectError, TimeoutException):
-        return 2
-    except Exception:
-        return 2
+    return generic_validate(url, process, headers=headers)
 
 
 if __name__ == "__main__":
--- a/user_scanner/creator/patreon.py
+++ b/user_scanner/creator/patreon.py
@@ -8,12 +8,6 @@ def validate_patreon(user):
 
 
 if __name__ == "__main__":
-    try:
-        import httpx
-    except ImportError:
-        print("Error: 'httpx' library is not installed.")
-        exit()
-
     user = input("Username?: ").strip()
     result = validate_patreon(user)
 
--- a/user_scanner/creator/producthunt.py
+++ b/user_scanner/creator/producthunt.py
@@ -1,5 +1,4 @@
-import httpx
-from httpx import ConnectError, TimeoutException
+from user_scanner.core.orchestrator import status_validate
 
 
 def validate_producthunt(user):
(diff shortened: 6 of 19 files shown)


kaifcodec:
@VamatoHD This is what I was thinking about to implement instead of silently failing and just error it would return the reason of every error. So that every user can even know, what criteria they should follow to get their desired username on those specific sites.

 As discussed in #60 

Absolutely a great POC at right time.

VamatoHD:
Moved the `Result` to its own file to improve readability.
The result can now store the `Exception` and I'm still working on converting this exception into a string. While a simple `str()` works, some exceptions wouldn't be understood by the user (like `[Errno 11001] getaddrinfo failed`).

@kaifcodec if you had any other idea or path in mind, feel free to comment it.

kaifcodec:
@VamatoHD I think you are doing well and the changes looks good also the `humanize_exception()` function that is handling the technical exception messages and let user know what exactly happening.

Anyways what I suggest is that for any other left (unknown) exception cases, just a message about report it on github by opening an issue will be enough.

Keep going by your own ideas, it will be fine.

VamatoHD:
@kaifcodec we should just continue to expand the message as we stumble upon them.
Also, how would verbose be implemented? What were your initial thoughts?

kaifcodec:
Yeah, for the verbose mode my initial thoughts were like, It would show the response body for the sites which will return errors.
I thought to implement it like if the `args.verbose` is true it would be passed to the worker functions as true and then it will show the response body of the errors but later I got that some sites return huge HTML, js response that will make no sense to show. So kept it paused.

If you have anything in mind to implement for that verbose after this addition you can do it though.

VamatoHD:
Maybe verbose should be the theme for another PR. My idea was that `verbose=True` would display the call stack; this means it would display when a function is executed, and `generic_validate`, `status_validate` and `generate_permutations` would display its results. But this idea may be further developed.

For now, I'll continue to develop this PR, improving `Result`.

VamatoHD:
I improved the Roblox module by adding better error messages and swapped the `int` for `Result` in most modules.

There are some modules that don't seem to work, so I didn't implement `Result` for them as of now. I'm going to focus on patching them.

kaifcodec:
@VamatoHD I tested it locally and your new class is doing its work properly but there are some unhandled Exceptions as you can see in the below output,

```bash
 Checking username: .kh._u.

== Community SITES ==
  [✔] Coderlegion (.kh._u.): Available

== Creator SITES ==
  [✔] Devto (.kh._u.): Available
  [✔] Hashnode (.kh._u.): Available
  [!] Itch.io (.kh._u.): Error (UnicodeError: Label empty or too long)
  [✔] Kaggle (.kh._u.): Available
  [✔] Medium (.kh._u.): Available
  [✔] Patreon (.kh._u.): Available
  [✔] Producthunt (.kh._u.): Available

== Dev SITES ==
  [✔] Codeberg (.kh._u.): Available
  [✔] Cratesio (.kh._u.): Available
  [!] Dockerhub (.kh._u.): Error (Status didn't match. Report this on Github.)
  [✘] Github (.kh._u.): Taken
  [!] Gitlab (.kh._u.): Error (Invalid status code)
  [✔] Launchpad (.kh._u.): Available
  [!] Npmjs (.kh._u.): Error
  [✔] Replit (.kh._u.): Available

== Donation SITES ==
  [✔] Buymeacoffee (.kh._u.): Available
  [✔] Liberapay (.kh._u.): Available

== Gaming SITES ==
  [✘] Chess.com (.kh._u.): Taken
  [✔] Minecraft (.kh._u.): Available
  [!] Monkeytype (.kh._u.): Error (Invalid status code)
  [!] Osu (.kh._u.): Error (Status didn't match. Report this on Github.)
  [!] Roblox (.kh._u.): Error (Too many requests)
  [!] Steam (.kh._u.): Error (Invalid status code)

== Social SITES ==
  [!] Bluesky (.kh._u.): Error (Invalid username)
  [✔] Discord (.kh._u.): Available
  [✔] Instagram (.kh._u.): Available
  [✘] Mastodon (.kh._u.): Taken
  [✔] Pinterest (.kh._u.): Available
  [✔] Reddit (.kh._u.): Available
  [✔] Snapchat (.kh._u.): Available
  [✔] Telegram (.kh._u.): Available
  [✔] Threads (.kh._u.): Available
  [!] X (Twitter) (.kh._u.): Error (
        X says: Your username can only contain letters, numbers and '_')
  [✔] Youtube (.kh._u.): Available

```
Now it's time for us to test those modules seperately, like `osu`, `monkeytype`, `dockerhub` etc. basically those who are throwing unhandled Exceptions and suggesting to report it on github we should fix those before merging this PR.

---

Anyways `Github` returns `422` for all errors, I will fix this and implement the Result class.

---
Edit: I fixed a minor issue in `X` module.

VamatoHD:
@kaifcodec The changes in the GitHub module look solid!

Also, I need to check why some websites always return forbidden; it must be something about my internet. Maybe we could implement the browser approach where it first tries to use `validate_<site>` and if it returns forbidden, the browser is executed, loads the website, and checks.

kaifcodec:
@VamatoHD Do you mean headless browsers, with selenium/playwright like tools?

VamatoHD:
@kaifcodec Yes! It was discussed in #31.

kaifcodec:
But I don't think we should implement it for now, cause its not a light weight tool, The `chromium`, `firefox` setup easily takes uto 600MB to 1GB (considering the cache, post installation).

Anyways on which websites you got `forbidden` errors?

VamatoHD:
The modules `medium`, `producthunt`, `npmjs` and `youtube` always return forbidden, I've even tried different headers.

The headless browser would be an **optional feature**, maybe activated by a flag. Even if the overhead is big, it would be worth it.

kaifcodec:
I figured it out, those URLs are showing `forbidden` only on your side because the API block is region/IP-based. If you try with a VPN it should work, and you can also verify by opening the links directly in a browser.

About `feat: headless`: it must stay completely optional. Before enabling it, we need to clearly document what it requires and how it will be used. We also need to make sure it won’t break due to architecture differences across systems.

But before we jump to that feature, we should finish the current drafted PR.  
Do you think it’s ready to merge now, or are you planning to push a few more changes?

VamatoHD:
@kaifcodec I haven't implemented the `Result` for the modules that I mentioned.
Also in the `Result.from_number` there is a mistake in `return cls(status,  reason if status == Status.TAKEN else None)` where it should be `        return cls(status,  reason if status == Status.Error else None)`.
Perhaps you should implement these before merging.

kaifcodec:
I am quite busy right now, but I would try to fix those issues.

VamatoHD:
@kaifcodec It just needs more testing.
If you don't have anything else to add, I think it's ready to be merged.

kaifcodec:
@VamatoHD It would be better if you  take a look at #76 and add the `Result` for that, cause the PR was made before this merge so maybe we have to edit it by ourselves.

PR #64, opened 2025-12-04
Implement wildcard-based username permutations with limit support

ganeshnikhil:
Implement wildcard-based username permutations

- Added `--permute` and `--stop` arguments to generate limited permutations
- Updated scanning modules to run for each generated username
- Display username along with site availability status
Screenshot 2025-12-04 at 11 34 50 AM

PR #67, opened 2025-12-04
Generalize categories

VamatoHD:
Instead of having to manually insert each category in `__main__.py` and in `orchestator.py` when created, I made it so they are automatically loaded:

- Created `load_categories() -> Dict[str, Path]` that returns a dictionary of the names of each category with its path;
- Changed `load_modules(category_path: Path)` to accept the category path instead;

### DIFF
--- a/user_scanner/__main__.py
+++ b/user_scanner/__main__.py
@@ -1,38 +1,20 @@
 import argparse
 import time
 import re
-from user_scanner.core.orchestrator import run_checks, load_modules , generate_permutations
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
@VamatoHD This looks like a solid implementation, should save us the headache of manually adding new category names every time we add one in the future.

If you have some time and are interested, you could also open a PR to update the `README.md` with the latest features and usage commands from PR #64.

PR #69, opened 2025-12-05
Update README Wildcard-based username permutations

ganeshnikhil:


kaifcodec:
@ganeshnikhil Nice approach, but it doesn’t fully cover the usage commands or how to use the wildcard username checks. The usage details should be in a separate point with a bash code block showing how the `-p` and `-s` flags work.

kaifcodec (approved):
@ganeshnikhil Now this one looks good, thanks for the update.

PR #76, opened 2025-12-08
added soundcloud username validator module

avramidis123:
Added a SoundCloud username checker. It's now in the social folder with the other platforms.

How it works:
- First checks if the page returns 404 (that means the username is free)
- If it's 200, it looks inside the page to see if it's a real profile or just a generic page
- Tries to be careful - if it's not sure, it says "available" so you can check manually
If you want me to change anything just let me know.


kaifcodec:
@avramidis123 It looks good! Just one question, does SoundCloud enforce any username restrictions (e.g., certain special characters not allowed, or characters like `_` and `.` not permitted at the beginning)?  
Also, does the API return a different response when the username format itself is invalid?

avramidis123:
From what i found testing soundcloud has username rules but their website seems to handle invalid usernames gracefully (returns 200 or 404, not a validation error). Characters like `_` and `.` at the beginning like (username` ,`.test`), SoundCloud returned 404 or a generic 200 page, not a specific "invalid format" error. This current module doesn't do pre-validation but i could add some if you wanted

kaifcodec:
@avramidis123 But I think in that case it might be better to use the private API that SoundCloud calls during the signup flow to check username availability. If you can, take a look and see whether that endpoint is accessible. 
Anyways if it requires authentication/cookies, then the current approach is fine.

avramidis123:
I checked - there's a private API endpoint but it requires authentication (client_id, possibly cookies). 
The current approach is simpler and doesn't need auth, similar to how other modules work.
Do you want me to try implementing the API version,or keep the current simpler version?

kaifcodec:
If that API requires authentication/cookies, then we should stick to this simpler approach for now.

We’ll take a little time before merging since the structure needs to be updated again after PR #63 was merged recently.

In the meantime feel free to push more commits, or go through #63 to understand the new structure if you prefer.

VamatoHD:
@kaifcodec I did change to the newer version with the `Result`; however, I don't understand why GitHub Desktop, while committing the new change, decided to include the commits from #63. Did I do something wrong?
I can do a reset and bring it to how it was.

kaifcodec:
@VamatoHD I think you didn’t do anything wrong, GitHub Desktop just merged main into your branch instead of rebasing, so the commits from #63 got included automatically.

If you can reset the branch back to the state where only your SoundCloud commit was present, that would clean up the PR history and make it easier to review.

After the reset, just push with `--force` so the PR updates.

If you are unable to do so, we must have to tell @avramidis123 to re-fork and open the PR again.

VamatoHD:
@kaifcodec I tried to do a `git reset --hard --force` but I don't have permission to write to his branch.

kaifcodec:
@VamatoHD Yeah, it supposed to be like that, and the issue happened just because his main branch was behind our main branch because we merged #63 in between this PR.

So it would be better if @avramidis123 re-fork this repository and open the PR again.

PR #80, opened 2025-12-09
Add StackOverflow username validator

Fort503:
## Title

Add StackOverflow username validator

## Description

This Pull Request adds a new validator for StackOverflow, located in `community/stackoverflow.py`

The StackOverflow user search endpoint:

https://stackoverflow.com/users/filter?search=

<username>

always returns HTTP 200, regardless of whether the user exists or not. Therefore, the validator uses `generic_validate()` and inspects the HTML response body.

The detection logic follows these rules:

- If the response contains the text `"No users matched your search."` → the username is available.  
- If the response contains an exact-match HTML pattern: `>{username}<` → the username is taken.  
- If neither condition is met, the username is treated as available, which also avoids false positives from partial matches.

### DIFF
--- a/user_scanner/community/stackoverflow.py
+++ b/user_scanner/community/stackoverflow.py
@@ -32,4 +32,4 @@ if __name__ == "__main__":
         print("Unavailable!")
     else:
         msg = result.get_reason()
-        print("Error occurred!" + (f" {msg}" if msg else ""))
+        print("Error occurred!" + msg)


user_scanner/community/stackoverflow.py
```diff
@@ -0,0 +1,35 @@
...
+    user = input("Username?: ").strip()
+    result = validate_stackoverflow(user)
+
+    if result == Result.available():
+        print("Available!")
+    elif result == Result.taken():
+        print("Unavailable!")
+    else:
+        msg = result.get_reason()
+        print("Error occurred!" + (f" {msg}" if msg else ""))
```
VamatoHD:
```suggestion
        print("Error occurred!" + msg)
```
Since `result.get_reason()` returns `""` if the Result doesn't have a reason, this check is useless.

user_scanner/community/stackoverflow.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def process(response):
+        if response.status_code == 200:
+            text = response.text
+
+            if "No users matched your search." in text:
+                return Result.available()
+
+            pattern = f'>{user}<'
+            if pattern in text:
```
VamatoHD: Does lowering both `pattern` and `text` improve result accuracy?

```suggestion
            if pattern.lower() in text.lower():
```

VamatoHD (commented):
This implementation looks solid! I ran some tests, and it appears to work perfectly.
I'll run some more tests and hopefully approve.

Fort503 (commented):
nice

Fort503 (commented):
nice

kaifcodec:
Wait @Fort503 Why you closed the PR? I think you misunderstood, your PR is perfectly fine we were just discussing about changes.

Fort503:
Oh, alright, alright, thanks.

kaifcodec:
@VamatoHD Do you want to make any changes? Or else I am going to merge it just let me know.

PR #99, opened 2025-12-12
fix: Updated headers and implementation of Result class

kaifcodec:
- Updated headers to mobile version for better trust
- Implemented Result class for better result comparisons.

### DIFF
--- a/user_scanner/creator/producthunt.py
+++ b/user_scanner/creator/producthunt.py
@@ -1,25 +1,47 @@
-from user_scanner.core.orchestrator import status_validate
+from user_scanner.core.orchestrator import status_validate, Result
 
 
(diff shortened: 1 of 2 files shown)


user_scanner/creator/producthunt.py
```diff
@@ -1,25 +1,47 @@
...
     }
 
-    status_validate(url, 404, 200, headers=headers, follow_redirects=True)
+    return status_validate(url, 404, 200, headers=headers)
+
 
 if __name__ == "__main__":
     user = input("Username?: ").strip()
-    result = validate_producthunt(user)
+    result = validate_youtube(user)
```
VamatoHD: Shouldn't this be `producthunt`?

Issue #75, opened 2025-12-08
Feat: Structured Output (JSON & CSV)


kaifcodec:
### Feature Request: Structured Output (JSON & CSV)

- **Goal**
Implement command-line arguments to output scan results in machine-readable JSON and CSV formats for scripting and automation.

- **Motivation**
Current text output is hard to parse. Structured output increases utility for power users and makes the tool easier to integrate into external scripts.

- **Implementation Plan**:

1.  **New Command-Line Flag:** Add an optional flag, `-o` or `--output-format`, accepting values: `json`, `csv`, or `console` (default).

2.  **JSON Format (`-o json`):**
    * Output a valid JSON array of objects to `stdout`.
    * Each object must include fields: `site_name`, `username`, `available` (boolean), and `url`.

3.  **CSV Format (`-o csv`):**

    * Output valid CSV to `stdout`.
    * The first line must be a header row (e.g., `site_name,username,available,url`).
    * Subsequent lines contain data.

4.  **Integration:** Ensure the threading mechanism correctly collects all results before the final formatting step.

**To Do Checklist**
- [x] Implement argument parsing for `-o`/`--output-format`.
- [x] Develop logic to format results into a JSON string.
- [x] Develop logic to format results into a CSV string (including headers).
- [x] Update documentation (README) with the new output usage.

VamatoHD:
@kaifcodec I can take this. I may even manage to implement this as soon as `Result` is merged.

kaifcodec:
@VamatoHD That will be fine then, I am also thinking about Merging the #63 PR, I had recently commited a small change about the error you mentioned, was that what you were talking about?

VamatoHD:
@kaifcodec you just need to implement `Result` for the modules that I can't test (`medium`, `npmjs` ... ) and then you can merge. The other error I mentioned should be fixed, even though that function wasn't used.

kaifcodec:
Ok, anyways lets keep the discussion here focused about this issue.

VamatoHD:
@kaifcodec There is a problem: Since the print takes place in `orchestrator.py`, values like the url need to be returned from each module. The only solution I can think of is for the `Result` to store it. Since most use `generic_validate` and `status_validate`, it should be simple to implement.

Also, it's probably best to have a different file have all print templates, since they take up too much space.

kaifcodec:
[...]

PR #100, opened 2025-12-12
Json and csv output

VamatoHD:
Following issue #75, this PR contains the start of its implementation. I'm aware that, even though **I haven't implemented CSV yet**, much has changed.

1. Made `printer.py` which contains a class `Printer` that handles the printing. It is passed to almost every function in `orchestrator.py` so they can handle printing.
2. New parameter `last` needed to remove the last comma in JSON output format.
3. Removes delay after the last printing statement.

@kaifcodec If you have any suggestions, feel free to mention them.

To Do:
- [x] Implement CSV.
- [x] Change the output's indentation to 2 spaces instead of `\t`.
- [x] Remove duplicated logic in `__main__.py`.
- [x] Clean up code, especially variable naming.
- [x] Update README.md.
- [x] Testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,16 @@ Perfect for finding a **unique username** across GitHub, Twitter, Reddit, Instag
 
 ### Features
 
-- ✅ Check usernames across **social networks**, **developer platforms**, and **creator communities**.
-- ✅ Clear **Available / Taken / Error** output for each platform.
+- ✅ Check usernames across **social networks**, **developer platforms**, and **creator communities**
+- ✅ Clear **Available / Taken / Error** output for each platform
 - ✅ Robust error handling: It prints the exact reason (e.g. Cannot use underscores, hyphens at the start/end)
-- ✅ Fully modular: add new platform modules easily.
+- ✅ Fully modular: add new platform modules easily
 - ✅ Wildcard-based username permutations for automatic variation generation using provided suffix
+- ✅ Selection of results format (e.g. json, csv, console (default))
+- ✅ Get the scanning results in preferred format (json/csv) in specified output file (suitable for power users)
 - ✅ Command-line interface ready: works directly after `pip install`
-- ✅ Can be used as username OSINT tool.
-- ✅ Very low and lightweight dependencies, can be run on any machine.
+- ✅ Can be used as username OSINT tool
+- ✅ Very low and lightweight dependencies, can be run on any machine
 ---
 
 ### Installation
@@ -57,7 +59,7 @@ Also, the output file and format can be specified: <br>
 user-scanner -u <username> -f console #Default format
 user-scanner -u <username> -f csv
 user-scanner -u <username> -f json
-user-scanner -u <username> -o <output-file>
+user-scanner -u <username> -f <format> -o <output-file>
 ```
 
 Generate multiple username variations by appending a suffix:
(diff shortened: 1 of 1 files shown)


kaifcodec:
@VamatoHD I think everything looks good so far, including the `Printer` class. It feels reusable and could be handy later for different modes as well.
Take your time and keep going, your approach looks solid.

VamatoHD:
@kaifcodec I think I'm finished, unless you have anything else to add. I just think it needs more testing (maybe implement automatic unit testing in another PR).

To sum up:
1. Added output format for JSON and CSV.
1.1 This applies for the scan and for the list command.
2. Cleaned up repeated code in main.
2.1 Created a function `find_modules(name: str)` to avoid nested loops.
2.2 Made a function to run for all usernames.

There could be more cleanup, like the `is_last` logic being repeated everywhere (a `utils.py` file could be helpful).

kaifcodec:
@VamatoHD Got it. I’ll review it and see if anything needs changing, but overall it looks good to me.

And sure, if you think moving repeated logic into `utils.py` makes things cleaner, go for it.

About unit tests, I was also thinking of setting up GitHub Actions to automatically test different command combinations, which could help catch issues early.

VamatoHD:
The repeated logic of `is_last`, used for making the JSON work, isn't a big problem right now. I will improve it later. 
I will probably focus on implementing unit tests (or help you, since I've never made them before).

kaifcodec:
Got it, that makes sense. The `is_last` repetition isn’t a big issue for now, we can clean it up later if needed.

By the way, for testing, are you thinking of using `pytest` with `parametrize`, or doing more direct matrix-style tests in a GitHub Actions workflow? Let me know what you prefer.

VamatoHD:
To be honest, I have no idea.
Maybe each function could be checked individually and then each module is somehow checked.
Probably we can test the output using various random generated names, this way the chances of it being valid are high.

kaifcodec:
@VamatoHD ok, no issues I will open a PR about it maybe a draft pr if needed.

kaifcodec:
@VamatoHD I see what you’ve built. Right now it outputs the results in JSON and CSV format directly in the terminal, but issue #75 was intended to save those outputs into separate JSON or CSV files (based on what the user chooses) inside a new `results` directory.

The issue description was a bit unclear, so no worries there. If you’re up for adjusting it to write the output to files instead, I don’t think it should take long since most of the core logic is already in place.

- The console output should stay the same as it is now (default behavior).

- The main change needed is after all checks are finished: take the full result set and write it to a new JSON file (using `json.dump`) or a CSV file, based on the selected format. You can also add a flag for output path selection, with the default being the `results/` directory where the tool is run.

- You don’t need to remove anything existing, like the new `Printer` class. It’s better to keep it as is, since it helps users see what’s being written into the JSON/CSV files.

No rush at all, take your time.

VamatoHD:
@kaifcodec No worries! I thought `stdout` was the terminal, and users would pipeline it. 
I'll get into fixing it.

VamatoHD:
@kaifcodec I think I didn't miss anything, but it's better to check.
I made it so `Result` now stores everything, like URL, site_name… and added a new argument `-o` for specifying the output file.

kaifcodec:
@VamatoHD, this looks perfect now. There was a tiny overlook in the `Printer` class: when the format was `console` (default), the `get_start()` method was implicitly returning `None` because it was never in an `else` block. I've fixed this to return an empty string.

By the way, I think the `url` field doesn't make sense here. Since the tool is primarily for availability checks, what use will a user have for hardcoded API URLs? What are your thoughts on this?

VamatoHD:
@kaifcodec When I was implementing it, I didn't really think of use cases for it. Perhaps it's better to remove it. Don't forget to remove it in `hashnode.py` and in `discord.py`.

kaifcodec:
No worries, I’ll fix those. Along with that, I’ll do a few cleanups and small changes and try to get it merged by tomorrow if possible. I’m a bit busy these days, though.

json-hunter07 (approved):

Issue #90, opened 2025-12-11
Add username support for TikTok


TedTschopp:
### TikTok username validation

**Category:** `social`

**Goal:** Add a validator for TikTok handles.

**Suggested URL format (to verify):**
- `https://www.tiktok.com/@{username}`

**Likely logic:**
- `200` → username exists (Taken)
- `404` or a dedicated error page → username not found (Available)
- Anything else → Error

**Implementation notes:**
- Add `social/tiktok.py` with `validate_tiktok(username: str)`.
- Use `status_validate()` if 404/200 is reliable; otherwise use `generic_validate()` and inspect `response.text`.
- Consider adding basic User-Agent headers similar to other social validators.

PR #101, opened 2025-12-14
Added bitbucket and sourceforge

### LINKED ISSUE #85: Add username support for Bitbucket
### Bitbucket username validation

**Category:** `dev`

**Goal:** Add a validator module for Bitbucket user profiles.

**Suggested URL format:**
- Candidate profile URL: `https://bitbucket.org/{username}/`

**Logic (to confirm while implementing):**
- `200` → username exists (Taken)
- `404` → username not found (Available)
- Anything else → Error / unknown

**Implementation notes:**
- Add `dev/bitbucket.py`.
- Expose a `validate_bitbucket(username: str)` function.
- Prefer `status_validate()` from `core/orchestrator.py` if the status codes are reliable; otherwise use `generic_validate()` with a `process(response)` function.
- Follow patterns in existing dev modules like `github.py`, `gitlab.py`.

**Nice to have:**
- Add a short reason message via `Result.error("...")` when Bitbucket returns an unexpected response.

### LINKED ISSUE #86: Add username support for SourceForge
### SourceForge username validation

**Category:** `dev`

**Goal:** Add a validator module for SourceForge user profiles.

**Suggested URL format (to verify):**
- Candidate profile URL: `https://sourceforge.net/u/{username}/`

**Logic (to confirm while implementing):**
- If distinct `200` vs `404` responses exist, use `status_validate()`.
- If all responses are `200`, inspect body using `generic_validate()` and `process(response)` to distinguish existing vs non-existing users.

**Implementation notes:**
- Add `dev/sourceforge.py`.
- Expose `validate_sourceforge(username: str)`.
- Mirror style from other dev modules and keep dependency footprint minimal.

VamatoHD:
Just a simple addition of the modules `bitbucket` and `sourceforge`, both using a simple status check.
Closes #85 and closes #86.

### DIFF
--- a/user_scanner/dev/bitbucket.py
+++ b/user_scanner/dev/bitbucket.py
@@ -1,7 +1,18 @@
-from user_scanner.core.orchestrator import status_validate
+import re
+from user_scanner.core.orchestrator import status_validate, Result
 
 
-def validate_bitbucket(user: str):
+def validate_bitbucket(user: str) -> Result:
+    if not (1 <= len(user) <= 30):
+        return Result.error("Length must be 1-30 characters.")
+
+    if not re.match(r'^[a-z0-9][a-z0-9_-]*$', user):
+
+        if re.search(r'[A-Z]', user):
+            return Result.error("Use lowercase letters only.")
+
+        return Result.error("Only use lowercase letters, numbers, hyphens, and underscores.")
+
     url = f"https://bitbucket.org/{user}/"
 
     return status_validate(url, 404, [200, 302], follow_redirects=True)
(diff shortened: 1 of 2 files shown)


kaifcodec:
I’d suggest adding a regex validation step before passing the username to the validator. If the username contains special characters that aren’t supported by Bitbucket or SourceForge, you can return it via `Result.error()` instead. This should help avoid confusion and make the behavior clearer.

kaifcodec:
@VamatoHD I implemented the regex validation, just go through the changes once.

VamatoHD:
@kaifcodec looks good to me!

kaifcodec:
@VamatoHD If you can take the issue #90 it would be better, check that out whenever you are free.

Issue #92, opened 2025-12-11
Add username support for LinkedIn


TedTschopp:
### LinkedIn vanity URL validation

**Category:** `social`

**Goal:** Check availability of LinkedIn public profile URLs (vanity names).

**Notes:**
- LinkedIn usernames typically appear in URLs like `https://www.linkedin.com/in/{username}/`.
- LinkedIn may return 999/redirects or require careful handling of headers.

**Implementation suggestions:**
- Investigate the behavior of `https://www.linkedin.com/in/{username}/` for existing vs non-existing users.
- If a clean 404 is not available, use `generic_validate()` and look for strong indicators in `response.text`.

**Implementation notes:**
- Add `social/linkedin.py` with `validate_linkedin(username: str)`.
- Document any rate limit / bot-detection quirks in comments.

PR #107, opened 2025-12-16
LinkedIn Username Validation

venkatesh2100:
- #92 

The validator:
Requests the public profile URL: https://www.linkedin.com/in/{username}/
Does not follow redirects to avoid triggering additional anti-bot checks
Common block responses include status 999, 403, and 429

- Error handling

Result.error("blocked (linkedin)") is returned intentionally when LinkedIn blocks or rate-limits the request.

### DIFF
--- a/user_scanner/social/linkedin.py
+++ b/user_scanner/social/linkedin.py
@@ -1,0 +1,69 @@
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
+
+
+def validate_linkedin(user: str) -> Result:
+    """
+    LinkedIn username validator.
+
+    Notes:
+    - LinkedIn aggressively blocks automated traffic.
+    - This validator is intentionally conservative.
+    - Blocked responses are treated as 'unknown'.
+    """
+
+    url = f"https://www.linkedin.com/in/{user}/"
+
+    headers = {
+        "User-Agent": (
+            "Mozilla/5.0 (X11; Linux x86_64) "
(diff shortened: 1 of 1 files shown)


kaifcodec:
@venkatesh2100 Thanks for the PR, it looks clean and well structured. However, when I tested it locally, it always returned a `999` status code, which is expected behavior from LinkedIn servers. LinkedIn requires a CSRF token in the request headers, and as a CLI tool it’s tricky to dynamically fetch that token and correctly craft the required headers. Because of this, the current approach won’t work reliably for us.

If we’re able to discover an undocumented or private API that LinkedIn uses for username availability checks, then this could be made to work. If you’re interested in exploring that (it’s a little advanced task), feel free to keep the PR open and push further commits if you find a solution.

Otherwise, you’re welcome to pick up any of the `good first issues` to work on in the meantime.

SylvainM98:
Hey @venkatesh2100, I found a way around the 999 block.

LinkedIn whitelists social media crawler bots (Twitter, Facebook, Bing) for link preview generation. Using the `Twitterbot/1.0` User-Agent, LinkedIn serves proper HTTP responses instead of 999: **200/301** for existing profiles, **404** for non-existing ones.

This makes the module a simple `status_validate()` call with no CSRF token or special headers needed.

I opened #264 with this approach. The module is ~15 lines and has been tested across existing profiles, non-existing ones, case sensitivity, hyphens, and edge cases.

PR #109, opened 2025-12-19
feat: auto-update feature

### LINKED ISSUE #108: Feature: Add auto update feature for new pypi release
### Description:
Add a runtime updater that checks the installed `user-scanner` version against the latest version on PyPI and optionally updates it.

### Details:
- Add updater logic under `utils/`
- Fetch latest version at runtime using `httpx`
  - GET: https://pypi.org/pypi/user-scanner/json
- Compare remote version with local version
- If mismatch, prompt user:
  - `y` → auto-update
  - `n` → skip
  - `d` → do not ask again
- Persist choice only for `d` in a new `config.py`
- Auto-update flow:
  - `pip uninstall user-scanner -y`
  - `pip install user-scanner`

### Notes:
- Fail silently if offline
- Do not block normal execution

kaifcodec:
Feature: Auto update Implementation
Closes #108

### DIFF
--- a/user_scanner/utils/updater_logic.py
+++ b/user_scanner/utils/updater_logic.py
@@ -59,4 +59,4 @@ def check_for_updates():
                 print(f"[{Y}*{X}] {R}Auto-update checks disabled.{X}")
 
     except Exception as e:
-        print(f"[!] Update check failed: {e}")
+        print(f"[{Y}!{X}] {R}Update check failed{X}: {e}")


kaifcodec:
@VamatoHD I’ve just implemented all of this fairly quickly. It still needs a bit more polish, which is why I opened it as a draft PR so we can work on it together and improve it gradually.

VamatoHD:
@kaifcodec I'll look into it as soon as possible, and, maybe, implement testing.

VamatoHD:
@kaifcodec I've replaced `config.py` with `config.json` since Python caches imports and could allow for better configurations in the future. Also, there was a typo in `update.py`.

From my tests, it appears to be working.

kaifcodec:
@VamatoHD Thanks for the review and changes in `config` file, I made some more changes mostly focused in colored outputs. 
Merging it!

PR #114, opened 2025-12-20
Introduce pytest-based testing and CI via GitHub Actions

VamatoHD:
The beginning of the implementation of tests. I've made some to test `Result` and `Printer`, but they need improving.
Also, I have no clue how to work with GitHub workflows, so @kaifcodec you must do it.

To Do:
- [x] Add GitHub workflows
- [x] More tests for the updating logic
- [x] Tests for `orchestrator.py`
  - [x] Test if the output is valid JSON and CSV
- [x] Tests for `utils.py`
- [x] Tests for `version.py`
- [x] Some automatic way to test each module

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama
+          pip install httpx colorama types-colorama
           pip install ruff mypy
 
       - name: Show installed linters


kaifcodec:
@VamatoHD Ok, no worries, I will look at it when i am free, maybe tomorrow.

kaifcodec:
@VamatoHD What are the changes in the last commit (d168a04)?

VamatoHD:
@kaifcodec Removed `AnyResult` type, since `Result` is fully integrated. Changed `monkeytype` and `telegram` return type to be `Result`. In `orchestrator.py`, the URL was still being set and `is_last` had type `int` when it's a `bool`.
This way, mypy doesn't error if it gets used to check the code before any PR.

kaifcodec:
Oh, got it.

Anyways I added many tests in the recent commits and they are still not ready so I will wait for it and will try to add a CI workflow next.

VamatoHD:
Also, in the new `test_orchestrator.py`, the line `assert any(p.startswith("user") and len(p) > len("user") for p in perms)` shouldn't be `assert all(p == "user" or (p.startswith("user") and len(p) > len("user")) for p in perms)`? To check if all permutations have the prefix of the initial username.

kaifcodec:
You're right, all() better matches the function’s contract.
The test enforces the invariant instead of just detecting one valid case.

I wrote all the codes fast and didn't test it for edge cases, that's why it needs more refinements.

kaifcodec:
@VamatoHD I merged this PR earlier to get a better sense of control, but everything looks fine now. I had to do a bit of trial and error since our codebase wasn’t very ruff-friendly, so I made the necessary fixes and it’s working well now.

Issue #115, opened 2025-12-21
Bug: Error in monkeytype and mastodon username checks


json-hunter07:
```bash
[!] Monkeytype (abd.ran12): Error (Invalid status code)
[!] Mastodon (abd.ran12): Error (ReadTimeout: The read operation timed out)

```
Instead of showing the reason it throws these errors only when speacial characters like "." Or others used in the username.

PR #116, opened 2025-12-22
Fix Monkeytype check failing on usernames with special characters

dxv3:
Fixes part of #115

Monkeytype username checks were returning a generic "Invalid status code"
when usernames contained unsupported characters (e.g. dots).

This change surfaces Monkeytype’s own validation error response instead,
providing a clear and accurate reason for the failure instead of a status error.

### DIFF
--- a/user_scanner/gaming/monkeytype.py
+++ b/user_scanner/gaming/monkeytype.py
@@ -1,9 +1,11 @@
 from user_scanner.core.orchestrator import generic_validate
 from user_scanner.core.result import Result
+from urllib.parse import quote
+
 
 def validate_monkeytype(user: str) -> int:
-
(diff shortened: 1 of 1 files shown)


dxv3:
1st pr kinda nervy

kaifcodec:
@dxv3 Thanks! This PR looks solid and fixes the issue in the right way.  
Instead of pre-validating usernames with regex, you’re correctly surfacing Monkeytype’s own `validationErrors` from the API, which is more reliable and future-proof if their rules change.  

If this is your first PR, it’s a good start and aligns well with the project’s direction.  
Feel free to look through the issues section or propose new ideas if you’d like to contribute further.  
Just make sure to review the contributing guidelines so you’re familiar with how the helper modules in `core` are intended to be used.  

Nice work, clean and well-thought-out fix.

PR #121, opened 2025-12-24
Improve roblox check

VamatoHD:
Since Roblox's API only allows 1 request every 1 minute and a half, I added a new method using `status_validate`. 
Still using the API since this method can have some edge cases (banned/terminated users).
I didn't include username prechecks since some usernames don't follow them due to [system errors](https://roblox.fandom.com/wiki/Username). For example, usernames having more than 20 characters.


kaifcodec (approved):
This one is better though.

PR #128, opened 2025-12-28
feat: add Battle.net username validator

### LINKED ISSUE #97: Add username support for Battle.net
### Battle.net (Blizzard) username / BattleTag validation

**Category:** `gaming`

**Goal:** Explore whether Battle.net usernames or BattleTags can be reasonably validated for availability.

**Notes:**
- Battle.net identities are often more complex (BattleTag#1234) and may not map cleanly to a simple public profile URL.
- This issue is exploratory: document findings and implement a checker only if a stable, public endpoint can be used.

**Suggested approach:**
- Investigate any public web profiles or APIs for Battle.net players.
- If a reliable URL exists, implement a validator using the existing helper functions.

**Implementation notes:**
- Add `gaming/battlenet.py`.
- Implement `validate_battlenet(username: str)` or similar, documenting any caveats.

anonymort:
## Summary

- Adds username availability checking for **Battle.net** via Overwatch player search API
- Uses `generic_validate` to check JSON response: empty array = available, non-empty = taken
- Includes BattleTag username pre-validation (3-12 chars, starts with letter, alphanumeric only)

## Implementation Details

**Endpoint:** `https://overwatch.blizzard.com/en-us/search/account-by-name/{username}`

**API behavior:**
- `200` with `[{...player data...}]` → Username taken
- `200` with `[]` → Username available

**File:** `user_scanner/user_scan/gaming/battlenet.py`

## Test Results

| Username | Expected | Result |
|----------|----------|--------|
| `Faker` | Taken | Taken |
| `Blizzard` | Taken | Taken |
| `zzqnoexist` | Available | Available |
| `ab` | Error (too short) | Error |
| `123test` | Error (starts with number) | Error |

## Caveats (documented in code)

- This checks **Overwatch profiles** specifically via the Blizzard player search
- A Battle.net account may exist without an Overwatch profile (and vice versa)
- BattleTag discriminators (#1234) are stripped if provided

Closes #97

### DIFF
--- a/user_scanner/user_scan/gaming/battlenet.py
+++ b/user_scanner/user_scan/gaming/battlenet.py
@@ -1,0 +1,65 @@
+import re
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
+
+
+def validate_battlenet(user: str) -> Result:
+    """
+    Check username availability on Battle.net via Overwatch player search.
+
+    Battle.net uses BattleTags (Username#1234) but this validator checks
+    if the username portion exists in the Overwatch player database.
+
+    Note: This checks Overwatch profiles specifically. A username may exist
+    on Battle.net but not have an Overwatch profile, or vice versa.
+
+    API behavior:
+        - Returns JSON array with player data if username exists
+        - Returns empty array [] if username not found
+    """
+    # BattleTag username rules: 3-12 chars, letters/numbers, one optional #
+    # For this validator, we strip any #1234 discriminator if present
+    username = user.split('#')[0]
+
+    if not (3 <= len(username) <= 12):
+        return Result.error("Length must be 3-12 characters")
+
+    if not re.match(r'^[a-zA-Z][a-zA-Z0-9]*$', username):
+        return Result.error("Must start with letter, only letters and numbers allowed")
+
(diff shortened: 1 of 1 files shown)


kaifcodec:
@anonymort Thanks for the PR. We’ll test it locally against edge cases, make any necessary adjustments, and then merge it.

json-hunter07:
I don't know whether it is needed or not because Battle.net doesn't use unique usernames, they assign a tag by themselves at the end to make it unique.

VamatoHD:
While Battle.net uses tags and allows users to have the same username, this could still be used to check if the username is unique or duplicated, rather than if it is "available" (which it is always in this case).

json-hunter07:
@VamatoHD Yeah you are right, but in that case it would be better to print a small messege about it, by passing it through `Result.available()`, `.taken()` both.

VamatoHD:
Hey @anonymort, could you rebase this PR onto the current `main`?
I'd like to implement the new `result`'s functionality into this module.

kaifcodec:
@VamatoHD Aren't you able to change the files, by using
```bash
gh pr checkout 128
```
?

VamatoHD:
No, because it needs the newer code from #142.

PR #129, opened 2025-12-29
feat: add Gumroad username validator.

atharva7-g:
## Summary

Added username availability for Gumroad at [gumroad.com](https://www.gumroad.com).
Uses `status_validate`. 
Verified username requirements (between 3 and 20 characters, must include at least one letter and may include only lower case letters and numbers) using the Gumroad website.
I have added a check for ASCII characters as well since the website doesn't seem to support non-ASCII characters.

## Details

Endpoint: `https://{user}.gumroad.com`

user_scanner/user_scan/creator/gumroad.py

## Test

| Username     | Expected                    | Result     |
|--------------|-----------------------------|------------|
| usernametaken | Taken | Taken |
| Blizzard     | Taken                       | Taken      |
| ab           | Error (too short)           | Error      |
| 123test      | Taken  | Taken |

This is my first time contributing. So, please let me know if I have missed anything.

### DIFF
--- a/user_scanner/user_scan/creator/gumroad.py
+++ b/user_scanner/user_scan/creator/gumroad.py
@@ -1,14 +1,9 @@
+import re
 from user_scanner.core.orchestrator import status_validate, Result
 
 def validate_gumroad(user: str) -> Result:
-    if not 3 <= len(user) <= 20:
-        return Result.error("Username must be between 3 and 20 characters.")
-
-    if user != user.lower():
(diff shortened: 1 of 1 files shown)


kaifcodec:
@atharva7-g Thanks for the PR! Great job on your first contribution
I tested it locally and it works perfectly. I just did a small refactor to use regex-based validation instead of repeated `if` blocks, mainly to keep the function cleaner and more concise.

PR #131, opened 2026-01-20
refactor(result.py): swap formatting and status display logic for ema…

json-hunter07:
Email-based scanning support:

**Key Changes:**
- **Improved Results:** Updated the `Result` class to show **Green** (Registered) and **Red** (Not Registered) specifically for email lookups.

### DIFF
--- a/user_scanner/core/result.py
+++ b/user_scanner/core/result.py
@@ -23,12 +23,10 @@ CSV_TEMPLATE = "{username},{category},{site_name},{status},{reason}"
 
 def humanize_exception(e: Exception) -> str:
     msg = str(e).lower()
-
     if "10054" in msg:
         return "Connection closed by remote server"
     if "11001" in msg:
         return "Could not resolve hostname"
-
     return str(e)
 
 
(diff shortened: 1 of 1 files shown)


VamatoHD:
@json-hunter07 I'm wondering why swapping the colors. I think the tool is more focused on finding emails/usernames that aren't in use. Therefore, green should be used when an email isn't being used (like as the username checker does). But you could have a good explanation for this change.

json-hunter07:
@VamatoHD You’re correct that for username scanning, **Green** signifies an available handle. However, I swapped the logic for the Email OSINT mode because the user intent is fundamentally different here.

Username scanning is typically about **Account Creation** (finding a free name), but Email OSINT is about **Information Gathering** (finding where a target already exists/which sites do target use). In this context:

* **Registered (Hit):** This is the "success" state for an investigator because it confirms a digital footprint. We use **Green** to highlight discovered data.
* **Not Registered (Miss):** This means no data was found, so it is marked **Red**.

I’ve kept the original color logic for usernames so that "handle-hunting" remains intuitive, but for OSINT, we want the 'hits' to stand out visually as the primary objective.

PR #135, opened 2026-01-21
Enforce check for default config.json

VamatoHD:
`config.json` can easily be modified (automatically) and accidentally commited it without the default values. There are two solutions:
1. Add a test to check if the configs are default (what this PR does).
2. Add a test to check if the config file doesn't exist (the `load_config` function theoretically creates a new one when executed).

If you prefer the second option, I can easily modify the test.

### DIFF
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -1,5 +1,10 @@
 from user_scanner.utils import updater_logic as ul
 
+def test_default_config():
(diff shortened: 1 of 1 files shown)


kaifcodec:
@VamatoHD This one is a good addition, and the option 1 is correct way to handle this case.

kaifcodec (approved):

PR #136, opened 2026-01-21
refactor(social/): better error handling to avoid false positive response

kaifcodec:
- This PR is opened for refactoring existing `email_scan` modules and improve the error handling
- Minimizing the false positive response cases.


json-hunter07:
I also refactored the `adult/` sites minimizing the risk of false positive responses and better error handling.

PR #137, opened 2026-01-22
add(facebook.py): add new new module in social/ category

kaifcodec:
- New facebook module in `email_scan/social/`
- Total of 2 consecutive requests to collect all the required cookies
- Validation through: `recovery page`
- is_silent: `True`


json-hunter07 (approved):
@kaifcodec It looks fine and the `if/else` conditions are correctly filtering out the risk of false positive responses as well!

PR #24, opened 2025-11-15
Added valorant user validation

jaduplansti:
# Added #13 (RIOT)

I had to do some digging, but i made it work.

PR #142, opened 2026-01-25
Change: Allow results to store reasons independently of their status

VamatoHD:
With this change, `result` always prints the reason independently of its status.
Like `result.error`, `result.available` and `result.taken` accept the reason as an argument. The check in `result.from_number` was also removed.
This change allows for the implementation discussed in #128 .


kaifcodec:
@VamatoHD this one was required especially with the new implementation of the email osint mode and even now the PRs #128 #24 can be completed.

PR #147, opened 2026-01-26
fix: correct 302 logic in mastodon

zahradm:


kaifcodec:
@zahradm I’ve tested the fix locally and it works as expected. Thanks for the contribution!
I also refactored the error message to align with our existing standard modules.

zahradm:
Thanks for testing it and for improving the error message. I’m glad the fix works well with the existing modules. Appreciate your help.

kaifcodec:
@json-hunter07 Your concern makes sense. For now, it should be fine to merge since there are no known edge cases at the moment. If anything unexpected shows up later, it can be handled with a follow-up fix.

PR #160, opened 2026-01-27
Add Replit email validation module

ShivamGupta22a:
Adds async Replit email validator using /data/user/exists endpoint.

Currently Replit returns HTTP 403 for automated clients; module reports this gracefully via Result.error.

Implements module under user_scanner/email_scan/dev/replit.py and exports validate_replit.

### DIFF
--- a/user_scanner/email_scan/dev/__init__.py
+++ b/user_scanner/email_scan/dev/__init__.py
@@ -1,3 +1,1 @@
-from .replit import validate_replit as validate_replit
 
-__all__ = ["validate_replit"]


kaifcodec:
@ShivamGupta22a Does it return 403 all the time?

ShivamGupta22a:
Yes during local testing it consistently returns 403 for automated requests, even after establishing cookies and setting XHR headers.

Manual browser requests succeed, but programmatic clients appear blocked (likely bot protection / WAF). That’s why the module treats 403 as Result.error rather than assuming availability.

If you’d like, I can also add a short docstring explaining this behavior.

kaifcodec:
So you should look for the response body when it returns 403, whether it is due to CAPTCHA block or a missing cookie related issue.

ShivamGupta22a:
I  inspected the response body  it returns a Cloudflare “Attention Required” HTML page, not JSON, which confirms this is a CAPTCHA / bot protection block rather than a missing-cookie issue.

kaifcodec:
I reviewed the code and I think there could be some authenticity token or something that the sites sometimes return in their HTML response when you hit the `...signup/` endpoint and that is later used in the request payload of the email validation step.

Like this below module: https://github.com/kaifcodec/user-scanner/blob/a3838da68f76f3a39bf9b07ae1d0595233d55cc6/user_scanner/email_scan/social/mastodon.py#L24-L25

ShivamGupta22a:
Good catch  I inspected the signup HTML but didn’t find an obvious csrf-token meta tag like Mastodon exposes. The response body I’m getting is already Cloudflare “Attention Required”, so I never reach Replit’s real signup page HTML.

I can try adding a token extraction step if Replit exposes one once Cloudflare is passed, but currently the initial GET is intercepted by Cloudflare before any app level tokens appear.

I’ll re-check for hidden inputs / meta tokens and update if I find anything.

kaifcodec:
> I inspected the response body it returns a Cloudflare “Attention Required” HTML page, not JSON, which confirms this is a CAPTCHA / bot protection block rather than a missing-cookie issue.


@ShivamGupta22a Got it, so it looks like Replit is blocking the request early using Cloudflare bot detection, even before the actual flow is reached.

So maybe its better for you to take other issues if you want to and leave this one open I would like to review it myself later.

ShivamGupta22a:
Got it  thanks for taking a look. I’ll leave this PR open and move on to other issues for now. Let me know if you’d like me to try anything else here.

kaifcodec:
@ShivamGupta22a Yeah,

Remember you can add any sites of your own choice to the tool its not  a rule that you must have to take sites from the issue section because it is an OSINT tool so it doesn't matter how niche the sites are. You have full freedom to open new PRs adding new sites even new categories.

Thanks for you efforts!

ShivamGupta22a:
Thanks appreciate the guidance. I’ll look into adding a few other sites and open new PRs.

kaifcodec:
@ShivamGupta22a
After further research, I found that the flow was overcomplicated by assuming a CSRF token was required to avoid blocking the POST request. In reality, the Replit email validation is quite straightforward it doesn't require any cookies or token for that successful POST request.
I’ve fully refactored the implementation accordingly. Test it by yourself once with both registered and unregistered emails and let me know.

json-hunter07:
@kaifcodec This fix is correctly implemented, replit doesn't require any authentication or session tokens for email availability checks.

PR #163, opened 2026-01-27
Added new username module: Vinted

VamatoHD:
This PR implements username checking for **Vinted**. Since there is no API, it does a simple regex on the search page to find all usernames.
I've made something different from the other modules: since the program gets a list of usernames given by the search, if the intended username isn't found, the best match is displayed.

Also, a new shopping category was added to better match the email's categories.


kaifcodec (approved):
@VamatoHD This one is working fine! Merging it.

PR #180, opened 2026-02-03
add 2 new modules to adult/

kaifcodec:
Uses the forget password endpoint 
- is_silent=False
   - sends a username reminder for `babestation.py` if exists


kaifcodec:
Locally tested for both registered and not registered account.

PR #183, opened 2026-02-03
fix(__main__.py): correctly append JSON and CSV outputs to existing files

kaifcodec:
- Fix JSON output to append to existing files while preserving valid structure
- Fix CSV output to append without breaking format
- Avoid invalid JSON caused by string-based appending
- Handle existing, missing, and empty output files safely
- Replace append-mode JSON writes with load/merge/dump flow
- Ensure consistent behavior across repeated runs
- PEP8 formatting

### DIFF
--- a/user_scanner/__main__.py
+++ b/user_scanner/__main__.py
@@ -1,7 +1,8 @@
 import argparse
-import time
-import sys
+import json
 import re
+import sys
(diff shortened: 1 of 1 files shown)


VamatoHD (approved):
LGTM

Issue #186, opened 2026-02-04
[BUG] Bitbucket always returns registered


VamatoHD:
## Description:
The module `email_scan/dev/bitbucket.py` appears to always be returning “Registered”, even with the most absurd emails or with personal emails that I'm sure aren't registered.

Also, I'm using the latest version of the code and have tested it on multiple devices.

## How to reproduce:
Simply execute the scan with any email.

kaifcodec:
@VamatoHD I reviewed the code, the request flow, and the response body, and came to the conclusion that the previous approach was wrong. Explained in : (#187).

From what I can see so far, I’m not able to find any endpoint that isn’t CAPTCHA-protected and that reliably tells us about email availability. So we may need to remove the Bitbucket module entirely.

PR #187, opened 2026-02-05
fix(bitbucket.py): correct inverted registration check logic

Eltigani-web:
Fixes issue #186
The bug was caused by inverted boolean handling: when implicitConsent was True, the code returned Result.available(), and when False it returned Result.taken(). That flips “registered” and “not registered,” making everything look registered. Swapping those two return values restores the correct mapping and matches the pattern used in other email scan modules.

### DIFF
--- a/user_scanner/email_scan/dev/bitbucket.py
+++ b/user_scanner/email_scan/dev/bitbucket.py
@@ -8,10 +8,10 @@ async def _check(email: str) -> Result:
             url = "https://id.atlassian.com/rest/marketing-consent/config"
             payload = {"email": email}
             headers = {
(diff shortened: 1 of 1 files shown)


kaifcodec:
@Eltigani-web Thanks for the PR. After reviewing it, I found that the endpoint currently being used for the email check isn’t actually meant for email registration validation, it’s used to inform the frontend about regional/legal context based on the request’s origin. So the module was effectively broken from the start since we had added it.

Because of that, I’ll likely remove the entire `bitbucket.py` module if we can’t find a reliable endpoint from Atlassian for email validation that isn’t protected by CAPTCHA.

Thanks for stepping in to try to fix this. If you’d like to look for alternative endpoints for email checks, you’re very welcome to do so. And if you want to take ownership of the  issue #186  let me know, I can assign it to you so we don’t work on it in parallel.

PR #192, opened 2026-02-06
fix(__main__.py) and add(m...p*.py/s...p*.py) two new modules in adult and module name normalization

kaifcodec:
1. Normalise the usage of `.` in `-m` flag by replacing `.` with `_`

2. Addition of two new email_scan modules in `adult/` category
      - made_p***.py
          - is_silent=False
          - uses forgot password 
      - superp***.py
          - is_silent=True


json-hunter07 (approved):
@kaifcodec That's a nice normalisation! LGTM.

PR #195, opened 2026-02-07
Add tests for engine.py

vmi98:
Added tests for all the functions in engine.py.
I also noticed that it uses asyncio.iscoroutinefunction, that is deprecated and planned  for removal in Python 3.16. (https://docs.python.org/3/deprecations/index.html). Maybe you should replace it with inspect.iscoroutinefunction() that they recommend to use instead.

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -22,6 +22,11 @@ from user_scanner.core.result import Result
 
 
 @pytest.fixture
+def anyio_backend():
+    return 'asyncio'
+
+
+@pytest.fixture
 def github_stub():
     module = ModuleType("github")
     module.__file__ = "/some/path/email_scan/dev/github.py"


kaifcodec (dismissed):
The tests are failing locally because anyio defaults to running a Trio loop alongside the asyncio one. Since our engine and AsyncMock are strictly tied to asyncio, the Trio run crashes with a RuntimeError.
Adding this fixture fixes the issue by locking the test environment to the asyncio backend

kaifcodec (approved):
Thanks for the updates! I’ve integrated the `inspect.iscoroutinefunction` fix and the architectural docstring as we discussed.
Now that the tests are passing with your latest commit, this is ready to go. For future reference, you can actually just click "Commit Suggestion" on my review to apply those fixes instantly and keep the history even tighter, but regardless, the end result looks solid.

Great work on this!

PR #226, opened 2026-02-20
fix(facebook.py): adapting regex search to the new changes of respons e pattern of facebook's /login enpoint

kaifcodec:
### Fix: Update Facebook Token Extraction

Improved regex to handle Facebook's shift from HTML inputs to JSON-embedded tokens.

* **LSD:** Added support for the new `["LSD", [], {"token": "..."}]` JSON format.
* **Jazoest:** Added fallback to capture values from script-based URL parameters.
* **Reliability:** Implemented multi-pattern matching to support both legacy and modern "Comet" page layouts.

### DIFF
--- a/user_scanner/email_scan/social/facebook.py
+++ b/user_scanner/email_scan/social/facebook.py
@@ -35,16 +35,27 @@ async def _check(email: str) -> Result:
                 'priority': "u=0, i"
             }
             res2 = await client.get(url2, params=params2, headers=headers2)
+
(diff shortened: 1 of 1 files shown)


kaifcodec:
I've updated the token extraction since Facebook keeps shifting between their modern JSON blobs and the old HTML input tags. Added some fallbacks to catch both patterns so we don't get caught off guard when they randomly swap response layouts.

PR #232, opened 2026-02-22
refactor: for user_scan hit means green (Found) and miss means Red (Not Found)

kaifcodec:
- Change the username scanning logic to OSINT
- Replace Available, Taken with Not Found, Found 
- Hit = Green and Miss = Red


json-hunter07 (approved):
LGTM, as we moved to OSINT tool, this one fits with the logic now!

PR #238, opened 2026-02-23
feat: add --only-found flag to filter output to found results only

### LINKED ISSUE #236: Feature: Add --only-found flag to filter console output for hits only
### Description
Currently, when performing a full scan or a category scan, the tool outputs every check, including sites where the username/email was **not found**.

We need a new flag, `--only-found`, that filters the console output to display only successful matches.

---
### Proposed Behavior
- **Flag:** `--only-found`
- **Scope:** This should apply to **Category scans (`-c`)** and **Full scans** (default). 
- **Exclusion:** This flag should be ignored when running a **Single Module scan (`-m`)**, as the user is explicitly looking for a specific result there.
---

#### Username Mode:
If `--only-found` is active, only sites with the status `[✔] Found` (Status.TAKEN) should be printed to the terminal.

#### Email Mode:
If `--only-found` is active, only sites with the status `[✔] Registered` (Status.TAKEN) should be printed to the terminal.

### Tasks
- [ ] Add `--only-found` to the `argparse` logic in `__main__.py`.
- [ ] Update `Result` class in `result.py` with a helper method (e.g., `is_found()`) to check if `self.status.value == 0`.
- [ ] Modify `orchestrator.py` (User) and `email_orchestrator.py` (Email) to wrap the `result.show()` or `print` calls in a conditional check.
- [ ] Ensure the flag is passed correctly through the asynchronous worker in `email_orchestrator.py`.

### Example Usage
```bash
# Should only show sites where 'target' has an account
user-scanner -e target@email.com --only-found

# Should show all sites in Social category where 'target' has an account
user-scanner -u target -c social --only-found

hacker4257:
## Summary

Adds an `--only-found` CLI flag that filters console output to only show sites where the username or email **was found** (i.e., `Status.TAKEN`).

When `--only-found` is passed:
- Sites where the user was **not found** (`Status.AVAILABLE`) are suppressed from output
- Error results (`Status.ERROR`) are also suppressed
- Only positive hits (`Status.TAKEN` / "Found" / "Registered") are printed
- File output (`--output`) still includes all results regardless of the flag

## Changes

- `user_scanner/__main__.py`: Added `--only-found` argparse argument, passed through to all scan functions
- `user_scanner/core/result.py`: Updated `show()` method to skip printing when `only_found=True` and result is not TAKEN
- `user_scanner/core/orchestrator.py`: Threaded `only_found` parameter through `run_user_module`, `run_user_category`, and `run_user_full`
- `user_scanner/core/email_orchestrator.py`: Threaded `only_found` parameter through all email scan functions

## Test plan

- [ ] Run `user-scanner -u <username> --only-found` and verify only found sites appear
- [ ] Run `user-scanner -u <username>` without the flag and verify all results appear (no regression)
- [ ] Run `user-scanner -e <email> --only-found` and verify only registered emails appear
- [ ] Run with `--only-found --output results.json` and verify the output file still contains all results
- [ ] Run with `--only-found -c social` and verify category-scoped filtering works
- [ ] Run with `--only-found -m github` and verify module-scoped filtering works

Closes #236

### DIFF
--- a/user_scanner/core/email_orchestrator.py
+++ b/user_scanner/core/email_orchestrator.py
@@ -12,7 +12,7 @@ from user_scanner.core.result import Result
 MAX_CONCURRENT_REQUESTS = 15
 
 
-async def _async_worker(module: ModuleType, email: str, sem: asyncio.Semaphore, show_url: bool = False) -> Result:
+async def _async_worker(module: ModuleType, email: str, sem: asyncio.Semaphore, show_url: bool = False, only_found: bool = False) -> Result:
     async with sem:
         module_name = module.__name__.split(".")[-1]
         func_name = f"validate_{module_name}"
@@ -27,7 +27,7 @@ async def _async_worker(module: ModuleType, email: str, sem: asyncio.Semaphore, 
 
         if not hasattr(module, func_name):
             return (
-                Result.error(f"Function {func_name} not found").update(**params).show(show_url=show_url)
+                Result.error(f"Function {func_name} not found").update(**params).show(show_url=show_url, only_found=only_found)
             )
 
         func = getattr(module, func_name)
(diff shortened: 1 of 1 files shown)


json-hunter07:
@hacker4257 This PR was good, why you closed it?

PR #257, opened 2026-03-02
Added Any.do and Firefox email existence modules

anubhavanonymous:
This PR adds two new email checking modules:

- Any.do
- Firefox

Details:
- Proper async handling using httpx
- Added correct headers and response parsing
- Tested with both existing and non-existing emails

Both services were verified to return consistent results.

Let me know if any adjustments are needed.

### DIFF
--- a/user_scanner/email_scan/other/firefox.py
+++ b/user_scanner/email_scan/other/firefox.py
@@ -1,29 +1,43 @@
 import httpx
 from user_scanner.core.result import Result
 
-
 async def _check(email: str) -> Result:
     url = "https://api.accounts.firefox.com/v1/account/status"
     show_url = "https://firefox.com"
 
-    async with httpx.AsyncClient(timeout=5.0) as client:
-        try:
-            response = await client.post(
-                url,
-                data={"email": email}
-            )
+    payload = {
+        "email": email
+    }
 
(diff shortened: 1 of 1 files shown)


kaifcodec:
@anubhavanonymous  

- Have you tested it with both registered and unregistered emails?  
- Is there any chance of frequent `429` or `403` responses?  
- Also, add proper `if` blocks to safely handle cases where the server returns `429` or `403`.  

You can follow the standard module structure from the recent addition here for reference: [mewe.py](https://github.com/kaifcodec/user-scanner/blob/main/user_scanner/email_scan/social/mewe.py).  

Make sure it aligns with the existing pattern used across the project.

Other than that, your additions look solid.

anubhavanonymous:
I've updated both modules to align fully with the structure used in mewe. py, including explicit handling for 403 and 429 responses, proper HTTP status validation, and timeout exception handling.

I tested again with both registered and unregistered emails and confirmed expected behavior.

Please let me know if any further refinements are needed.

anubhavanonymous:
Thank you for the review and for the detailed feedback throughout the process. I really appreciate the guidance.

I'll continue exploring additional platforms and make sure future modules strictly follow the project's structure and stability standards.

Looking forward to contributing more

PR #262, opened 2026-03-02
fix: add Netflix to LOUD_MODULES (sends verification code to target)

SylvainM98:
## Summary

- Add `netflix` to the `LOUD_MODULES["email"]` list in `helpers.py`
- The Netflix module uses a `CLCSWebInitSignup` GraphQL mutation that triggers a **real verification code** sent to the scanned email address
- This notifies the target that someone is probing their account, which is the exact behavior the loud module mechanism is designed to prevent

## How it was discovered

During testing, each scan triggered an unsolicited Netflix verification code to the target email. The module should be skipped by default (like `instagram`, `leetcode`, `polarsteps`, etc.) and only run when `--allow-loud` is explicitly passed.

## Test plan

- [x] `user-scanner -e test@example.com` → Netflix shows `[~] Skipped`
- [x] `user-scanner -e test@example.com --allow-loud` → Netflix runs normally
- [x] `user-scanner -e test@example.com -m netflix` → Netflix runs (specific module mode)

### DIFF
--- a/user_scanner/email_scan/entertainment/netflix.py
+++ b/user_scanner/email_scan/entertainment/netflix.py
@@ -13,8 +13,8 @@ async def _check(email: str) -> Result:
     }
 
     try:
-        async with httpx.AsyncClient(timeout=5.0, follow_redirects=True) as client:
-            await client.get("https://www.netflix.com/in/", headers=headers)
+        async with httpx.AsyncClient(timeout=10.0, follow_redirects=True) as client:
+            await client.get("https://www.netflix.com/", headers=headers)
 
             flwssn = client.cookies.get("flwssn")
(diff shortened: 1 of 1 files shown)


SylvainM98:
Added a second commit that fixes the Netflix module itself:

- Switch to LOGIN node (reduces target notifications compared to WELCOME)
- Fix locale from en-IN to en-US for global compatibility (the module was India-only before)
- Fix session URL from netflix.com/in/ to netflix.com/
- Simplify detection to a single reliable marker
- Add GraphQL error guard to prevent false negatives if the API changes
- Increase timeout from 5s to 10s for better reliability during concurrent scans

Ready for review.

PR #263, opened 2026-03-02
add: Facebook username validation (social category)

SylvainM98:
## Summary

- Adds Facebook username validator using the account recovery lookup endpoint
- Includes client-side pre-validation (Facebook username format rules)
- Does **not** trigger any notification to the target (silent module)

The previous approach (PR #46) checked `facebook.com/{username}` HTTP status codes, but Facebook now returns HTTP 400 for all profile URL requests regardless of whether the username exists. This implementation uses the same recovery endpoint as the existing email scanner module, adapted for synchronous username lookups.

### Pre-validation checks
- Length: 1–50 characters
- Characters: letters, numbers, periods only
- No numeric-only usernames
- No leading/trailing periods

### Tested with
- Known existing usernames → Found
- Non-existent usernames → Not Found
- Various edge cases (empty, special chars, numeric, too long) → Proper error messages

Supersedes #46.

## Test plan

- [ ] `user-scanner -u <existing_username> -m facebook` → `[✔] Found`
- [ ] `user-scanner -u <fake_username> -m facebook` → `[✘] Not Found`
- [ ] `user-scanner -u <username>` (full scan) → Facebook appears in social results
- [ ] Invalid usernames (`12345`, `user@name`, empty) → `[!] Error` with clear message

### DIFF
--- a/user_scanner/user_scan/social/facebook.py
+++ b/user_scanner/user_scan/social/facebook.py
@@ -1,0 +1,117 @@
+import re
+
+import httpx
+from user_scanner.core.result import Result
+
+
+def _check(user: str) -> Result:
+    show_url = "https://facebook.com"
+
+    if not (1 <= len(user) <= 50):
+        return Result.error("Length must be 1-50 characters")
+
+    if not re.match(r"^[a-zA-Z0-9.]+$", user):
+        return Result.error("Only letters, numbers and periods allowed")
+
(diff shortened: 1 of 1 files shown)


kaifcodec:
Closing this as there has been small issue (false positives) in this PR and the fixed part of this contribution has already been added in #323 .

PR #266, opened 2026-03-03
Add Riot ID validation module (gaming)

SylvainM98:
## Summary

Adds a Riot ID module that checks if a Riot ID (Name#Tag) is registered across Riot Games titles (Valorant, LoL, TFT, LoR). Supersedes #24.

## How it works

1. **Validates format**: Requires exactly one `#` with non-empty name and tag (3-5 chars)
2. **Queries riotid-lookup.com API**: `GET /api/lookup?riotId={Name}%23{Tag}`
3. **Parses response**: `{"isTaken": true/false}`

| Response | Result |
|---|---|
| `{"isTaken": true}` | Taken |
| `{"isTaken": false}` | Available |
| Status 400 | Error (invalid format) |

## Why this approach?

The original #24 and first version of this PR used tracker.gg + `curl_cffi` to bypass Cloudflare. After feedback from @kaifcodec about Termux/ARM64 compatibility, several alternatives were tested:

| Endpoint | Result |
|---|---|
| tracker.gg API | Blocked by Cloudflare TLS fingerprinting |
| HenrikDev API | Requires API key since v4 |
| Cloudscraper + tracker.gg | Cannot bypass TLS fingerprinting |
| Riot official API | Requires production API key |
| **riotid-lookup.com** | **Works with httpx, no auth, ~1.1s avg** |

## Changes

- `user_scanner/user_scan/gaming/riot_id.py` — new module using `generic_validate()` + httpx
- No new dependency (uses existing httpx)

## Test plan

- `python3 -m user_scanner -u 'TenZ#00005' -m riot_id` → Found
- `python3 -m user_scanner -u 'xyzfake99#99999' -m riot_id` → Available
- `python3 -m user_scanner -u 'TenZ' -m riot_id` → Error (missing tag)
- `python3 -m user_scanner -u 'Test#1' -m riot_id` → Error (tag too short)

### DIFF
--- a/user_scanner/user_scan/gaming/riot_id.py
+++ b/user_scanner/user_scan/gaming/riot_id.py
@@ -32,4 +32,11 @@ def validate_riot_id(user: str) -> Result:
 
         return Result.error(f"HTTP {response.status_code}")
 
-    return generic_validate(url, process, show_url=show_url)
+    headers = {
+        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 "
+                      "(KHTML, like Gecko) Chrome/141.0.0.0 Safari/537.36",
+        "Accept": "application/json",
+        "Accept-Encoding": "gzip, deflate",
+    }
+
+    return generic_validate(url, process, show_url=show_url, headers=headers)


VamatoHD:
I'm aware that it uses some API linked to Valorant; however, it is checking whether a Riot ID exists. If it does, the ID can be applied for other games like League of Legends.
Shouldn't it be marked as Riot ID instead?

kaifcodec:
However @VamatoHD I don’t have much familiarity with those games, so you can probably handle the logic in the way you think fits best.

While testing this module individually, I noticed it failing on some CPU architectures — specifically ARM64 environments like Termux on Android.
```bash
File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/curl_cffi/__version__.py", line 3, in <module>
  from .curl import Curl
File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/curl_cffi/curl.py", line 14, in <module>
  from ._wrapper import ffi, lib
ImportError: dlopen failed: library "libcurl-impersonate.so.4" not found: needed by /data/data/com.termux/files/usr/lib/python3.12/site-packages/curl_cffi/_wrapper.abi3.so in namespace (default)
```
I encountered this on Termux. The `libcurl-impersonate.so.4` missing error suggests that although the Python package installs, the underlying native dependency (`libcurl-impersonate`) isn’t available or compiled properly for the Android/Termux environment during `pip install`.

Because of this, if we include this dependency we should probably make the module optional. If the dependency isn’t available, the tool should skip this module gracefully instead of failing the scan.

This is especially important since we state in the README that the tool is **tested on Termux**.

SylvainM98:
Addressed both points:

**1. Renamed to Riot ID** (@VamatoHD)
- `valorant.py` → `riot_id.py` with `validate_riot_id()`
- Now covers all Riot Games titles (Valorant, LoL, TFT, LoR)

**2. Removed curl_cffi dependency** (@kaifcodec)
- Switched from tracker.gg to riotid-lookup.com API
- Uses httpx via `generic_validate()` — no new dependency, Termux compatible

**Alternatives tested:**

| Endpoint | Result |
|---|---|
| tracker.gg API | Blocked by Cloudflare TLS fingerprinting |
| HenrikDev API | Requires API key since v4 |
| Cloudscraper + tracker.gg | Cannot bypass TLS fingerprinting |
| Riot official API | Requires production API key |
| **riotid-lookup.com** | **Works with httpx, no auth, ~1.1s avg** |

**Performance comparison:**

| | curl_cffi + tracker.gg | httpx + riotid-lookup.com |
|---|---|---|
| Avg response | ~1.25s | ~1.13s |
| Native dep | Yes (breaks ARM64) | No |
| Scope | Valorant only | All Riot titles |

kaifcodec:
@SylvainM98 It would be better if you open a new PR with a single commit containing the final change for this module, instead of the current 5 commits with additions and deletions. This will help keep the commit history cleaner.

SylvainM98:
Replaced by #299 with a clean single commit.

PR #275, opened 2026-03-05
Add boot.dev to email and user scanner

aspectp:


kaifcodec:
@aspectp Thanks for adding the sites. However, could you open a new PR with clean commits? It looks like the `codecademy` module was accidentally overwritten.

That will help keep the history clean. I suggest re-forking the repository to avoid carrying over the previous commits when pushing again.

aspectp:
alright

PR #280, opened 2026-03-06
Feature: addition of --no-nsfw flag

VamatoHD:
Added a new flag, `--nsfw` or `--allow-nsfw` that allows the NSFW site to be shown; otherwise, they are skipped. It works with `-u`, `-e`, `-lu` and `-le`

For now it just checks if the module is inside a `adult/` folder and can be changed later if needed.

### DIFF
--- a/docs/FLAGS.md
+++ b/docs/FLAGS.md
@@ -9,6 +9,7 @@
 | `--only-found`              | Only show sites where the username/email was found          |
 | `--allow-loud`              | Enable scanning sites that may send emails/notifications    |
 | `--no-nsfw`                 | Disable NSFW site scanning                                  |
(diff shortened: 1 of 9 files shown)


kaifcodec:
@VamatoHD I’d suggest reversing the logic behind the `--allow-nsfw` flag. Skipping NSFW site scans by default doesn’t seem necessary, especially since they aren’t as problematic as the loud modules.

You could instead use a `--no-nsfw` flag, the opposite of what is currently implemented, which would skip those modules only when the flag is explicitly used.

Most users don’t really mind if adult sites appear in the console output, and disabling them by default would also mean adding extra flags just to perform a full scan.

PR #282, opened 2026-03-09
Preserve explicit proxy schemes when loading proxy files

shuofengzhang:
## What changed
- Updated `ProxyManager._load_proxies` to preserve any explicit proxy scheme instead of only preserving `http://`, `https://`, and `socks5://`.
- Kept the existing behavior of prepending `http://` when a proxy entry has no scheme.
- Added a regression test (`test_proxy_manager_preserves_explicit_schemes`) covering `socks4://`, `socks5h://`, and `https://` entries.

## Why
- The previous loader could mangle valid proxy strings that already included an explicit but non-whitelisted scheme (for example, `socks5h://...`) by prefixing `http://`.
- Preserving explicit schemes makes proxy-file handling more robust while keeping backward compatibility for bare `host:port` lines.

## Testing
- ✅ `source .venv/bin/activate && pytest -q` (126 passed)

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -269,6 +269,21 @@ socks5://socks-proxy.example.com:1080""")
     assert manager.proxies[1].startswith("socks5://")
 
 
+def test_proxy_manager_preserves_explicit_schemes(tmp_path):
+    proxy_file = tmp_path / "test_proxies.txt"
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
Thank you for the PR @shuofengzhang !

PR #287, opened 2026-03-10
Add Proton Mail username availability check

### LINKED ISSUE #284: Add username support for Proton Mail
## Proton Mail username validation

**Notes**

``curl -H "x-pm-appversion: web-mail@6.0.1.3" "https://account.proton.me/api/core/v4/users/available?Name=<<name>>%40proton.me&ParseDomain=1"``

will return ``"Code":12106`` if the username exists. It will return ``{"Code":1000}`` if the name does not exist.

The ``x-pm-appversion`` header seems to work for any version above 5, so a current version could be chosen or something very high that won't get outdated could be used. 6.0.1.3 was just a guess because I saw that ``web-mail@5.0.1.3`` was used previously and is now out of date.

It doesn't seem like each domain (protonmail.com, proton.me, pm.me, protonmail.ch) needs to be done separately but I am not sure.

Their signup page is a good place for more info: https://account.proton.me/mail/signup

YuF-9468:
Fixes #284

Adds a Proton Mail username availability validator using Proton's account availability endpoint (Code=12106 taken / Code=1000 available).

- New module: `user_scanner/user_scan/community/protonmail.py`
- Adds unit tests with mocked network responses.

Test plan:
- `uv run pytest -q tests/test_user_scan_protonmail.py`

### DIFF
--- a/user_scanner/user_scan/email/protonmail.py
+++ b/user_scanner/user_scan/email/protonmail.py
@@ -28,7 +28,7 @@ def validate_protonmail(user: str) -> Result:
 
     def process(response):
         if response.status_code not in [200, 409]:
-            return Result.error("Unexpected status code from Proton")
+            return Result.error(f"[{response.status_code}] Unexpected status code from Proton")
 
         try:
             data = response.json()


kaifcodec:
@YuF-9468 Thanks for the PR, but there are a few issues to address:

- It was added under the `community` folder, but it should be placed in a new `email` folder.
- A separate `test` file was added for this module, which isn’t required.
- The PR shows a different commit author (not necessarily a problem, but it would be good to know the reason).

If you’d like to fix this, it would probably be easiest to re-fork the repository and push the commits again cleanly.

YuF-9468:
Thanks for the detailed review — fixed as requested and force-pushed a clean commit.

Changes made:
1) Moved `protonmail.py` from `user_scanner/user_scan/community/` to new `user_scanner/user_scan/email/`.
2) Removed the separate `tests/test_user_scan_protonmail.py` file.
3) Rewrote the branch as a clean single commit.

Please take another look 🙏

kaifcodec:
@YuF-9468 the commit author appears to be different from your GitHub profile, could you clarify why?

YuF-9468:
Thanks for flagging this — you were absolutely right.

I have fixed the commit metadata and force-pushed:
- commit author: YuF-9468 <264538812+YuF-9468@users.noreply.github.com>

The previous mismatch was caused by my incorrect local noreply email configuration. Sorry for the confusion, and thanks again for catching it.

kaifcodec:
@YuF-9468 Can you test it on your side as well? I tested it and I'm getting a weird `409` response for some common usernames, but usernames I know should exist there seem to work fine.

Try testing with two usernames where you already know the results one that should return `found` and another that should return `not found` to verify the behavior.

YuF-9468:
Done — I retested on my side with one known existing username and one known non-existing username:

- `protonmail` (should be **found**)
  - API response: **HTTP 409**, `{"Code":12106,"Error":"Username already used"}`
  - Current module output: `Error` (`Unexpected status code from Proton`)

- `ocyfagxvgkz2fivo` (should be **not found / available**)
  - API response: **HTTP 200**, `{"Code":1000}`
  - Current module output: `Not Found / available` ✅

So this confirms the issue is specifically for existing usernames, where Proton returns 409 + Code 12106 and the current status-code gate marks it as error.

YuF-9468:
Supplement from my retest:

- The issue is reproducible specifically for **existing usernames**.
- Proton returns `HTTP 409` with `Code: 12106` for taken usernames.
- Current implementation treats non-200 as error, so taken usernames become `Error` instead of `Found`.

I can submit a follow-up patch to map `(409, 12106)` to `taken/found` while keeping `(200, 1000)` as `available/not found`.

YuF-9468:
Fixed and pushed.

I updated the Proton module to handle both expected API paths explicitly:
- `HTTP 409 + Code 12106` -> `taken/found`
- `HTTP 200 + Code 1000` -> `available/not found`

Anything else now returns a structured error with both status and code for easier debugging.

PR #295, opened 2026-03-11
Feat: enhance Result reporting and implement Vedantu learning module, refactor old modules with it

kaifcodec:
- Introduces a dynamic extra field to the Result class for capturing auxiliary data during scans.
- Updates the reporting core to support the new field across JSON, CSV, and debug string outputs.
- Adds localized terminal display logic for extra information using cyan formatting for better readability.
- Synchronizes the test suite to ensure full coverage of the new data fields and output formats.
- Implements the Vedantu learning module with support for capturing masked phone numbers via the extra field.

### DIFF
--- a/user_scanner/email_scan/shopping/vivino.py
+++ b/user_scanner/email_scan/shopping/vivino.py
@@ -41,7 +41,7 @@ async def _check(email: str) -> Result:
                 return Result.taken(url=show_url)
 
             if "account has been locked" in error_msg:
-                return Result.taken(url=show_url, extra="Acoount is locked by Vivino")
+                return Result.taken(url=show_url, extra="Account is locked by Vivino")
 
             return Result.error(f"Vivino Error: {error_msg}", url=show_url)
 


kaifcodec:
Yeah, that needs to be changed, it was just a demo.

I’m planning to show the `extra` details at the end of the scan, something like:

```bash
[i] Scan complete.

# Extra info (only if it gets any)

  Total hits: 1
```

This approach keeps the output cleaner. Also, I don’t expect too many extra details in most cases, especially in `email_scan` mode, where at best we might occasionally see phone numbers, which is still rare since most sites don’t disclose that information.

I mainly added this as a placeholder in case we move forward with extracting additional details from usernames in the future.

Let me know if you think something else would work better.

VamatoHD:
@kaifcodec I'll look into it. Also, I don't think showing it at the end is better.

kaifcodec:
@VamatoHD Ok, when it's ready let me know what you are going to do.

VamatoHD:
@kaifcodec After trying a couple of approaches, I really liked the following:
image
You can always revert it if it doesn't fit your liking.

kaifcodec:
@VamatoHD Yeah this approach is good, but we also need to adapt it later when we move forward for the deep scans in future to return the `data`s as tree.
But right now its good!

json-hunter07 (approved):

PR #301, opened 2026-03-17
fix(helpers.py): replace `made_porn` with `made.porn` in LOUD_MODULES

json-hunter07:
- Initially it was ignoring `made_porn` as no module found as `get_site_name()` function already coverts the `made_porn` to `made.porn`
- Now it correctly matches the module's name


kaifcodec (approved):
Good catch @json-hunter07 !

PR #304, opened 2026-03-20
Add cashapp and pastebin to user scanning

aspectp:
Tested on existing accounts


kaifcodec:
@aspectp Thank you for Opening this PR, I tested both of the modules.
I noticed that `cashapp` blocking the request via cloudflare error `403`.

However your `pastebin` module addition works fine. So it would be better to open a new PR only adding the `pastebin` module.

kaifcodec:
Closing this PR as there has been no response to the previous review comments, and the working part of this contribution has already been added in #315.

PR #315, opened 2026-04-01
Add pastebin support

Brumbelow:
- Added pastebin support from previous PR


kaifcodec (approved):
@Brumbelow Thank you for the contribution and for separating it from PR #304 as requested.

PR #316, opened 2026-04-02
fix: false positive and reddit validation logic to use json api

kaifcodec:
### **Changes**:
* Switched from HTML scraping to the `user/{user}/about.json` endpoint.
* Added handling for `404` status codes and JSON error payloads.
* Implemented `kind: t2` verification for accurate account detection.


json-hunter07 (approved):
This fix is really an upgrade! LGTM.

PR #323, opened 2026-04-13
fix: Added the required public validator validate_facebook(user: str) and aligned with CONTRIBUTING.md

Brumbelow:
- Kept the existing synchronous Facebook probe, but fixed it to return the repo’s supported Result types and attach url=... consistently at user_scanner/user_scan/social/facebook.py:9 and user_scanner/user_scan/social/facebook.py:122.

- Fixed the broken LSD token regex so group(1) is valid instead of risking a runtime failure during token extraction at user_scanner/user_scan/social/facebook.py:70

- checks pass


user_scanner/user_scan/social/facebook.py
```diff
@@ -0,0 +1,137 @@
...
+                "accept-language": "en-US,en;q=0.9",
+            }
+
+            params = {
+                "username": user,
+                "fb_dtsg": lsd,
+                "jazoest": jazoest,
+            }
+            response = client.get(
+                "https://www.facebook.com/ajax/username availability/",
```
kaifcodec: @Brumbelow I’m a bit skeptical about this URL, it doesn’t seem to be a valid endpoint. It’s currently returning a 404, which suggests there isn’t an actual API route like this.

It would be better to capture the real request flow during username checks and use the exact endpoint being called there.

Brumbelow: Yes looks like a hallucination in the previous PR. I think I found a usable route with just ```https://www.facebook.com/{user}```. I will update shortly

kaifcodec: @Brumbelow Yeah the previous PR looks like an AI generated guess.
Anyways after updating with your latest find, test it yourself once with both `taken` and `available` usernames

Brumbelow: Should be better now,  zuck -> Found and thisshouldnotexistzzzzzzzzzz2026 -> Not Found. 

lmk

kaifcodec:
@Brumbelow It looks good to me now. While testing, I noticed that for the first 6–7 attempts with a non-existent username it returned the `Unexpected response body` error, but after that it became consistent and has been stable since.

I’ll give it a bit more time and test a few edge cases, if everything looks good, it should be merged within a few hours.

If you’d like to contribute further, feel free to add support for other sites in both `email_scan` or `user_scan`.

kaifcodec (approved):
@Brumbelow Right now it looks good, I have deleted the `tests` you had added for Facebook seperately.
Merging it.

json-hunter07 (approved):
LGTM, @kaifcodec

PR #327, opened 2026-04-23
refactor(lovescape): update availability logic to infer email status via username validation

kaifcodec:
- Refined logic in `lovescape.py` to improve accuracy for email availability checks.
- Updated the error handling flow to correctly interpret "Username is already used" API responses.
- Leveraged the API's validation sequence: if the "email already used" check passes but the username check triggers, the email is confirmed as not registered.
- Enhanced reliability of the module by correctly mapping API error responses to `Result.available` status.


json-hunter07 (approved):
@kaifcodec Yeah it is now correctly handling the validation.
It seems like the site had changed their sign up logic.
Merging it!

PR #328, opened 2026-04-23
fix(instagram): replace broken web_profile_info with signup-attempt endpoint

anonymort:
## Summary

- **Bug:** Instagram's `web_profile_info` API endpoint now returns `401 Unauthorized` for all unauthenticated requests, causing every Instagram username scan to fail with `[401] Status didn't match`
- **Fix:** Switch to the signup-attempt endpoint (`web_create_ajax/attempt`), which reliably reports whether a username is taken without requiring authentication
- **Approach:** POST a registration attempt with just the username — Instagram's response includes a `username` error with code `username_invalid` or `username_is_taken` if the name is taken, and no username error if it's available. CSRF validation only checks that the header matches the cookie value, so a dummy token (`"0"`) works without an extra preflight request

## Test plan

- [x] Verified known taken usernames (`instagram`, `cristiano`, `bonniebop`) report **Found**
- [x] Verified non-existent usernames (`xyznonexistent99999`) report **Not Found**
- [x] Verified full CLI works end-to-end: `user-scanner -u <name> -m instagram`
- [x] Single HTTP request per check (no setup/teardown overhead)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/user_scanner/user_scan/social/instagram.py
+++ b/user_scanner/user_scan/social/instagram.py
@@ -1,34 +1,46 @@
+import httpx
+
 from user_scanner.core.helpers import get_random_user_agent
-from user_scanner.core.orchestrator import status_validate
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
 
 
 def validate_instagram(user):
-    url = "https://www.instagram.com/api/v1/users/web_profile_info/"
     show_url = f"https://www.instagram.com/{user}/"
-
-    params = {"username": user}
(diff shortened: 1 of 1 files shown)


anonymort:
Verified this fix works end-to-end — ran a bulk scan of 41 usernames and Instagram checks returned correct results for all of them (e.g. `bonniebop`, `bonniebean`, `missbonnie` all correctly reported as Found). The signup-attempt endpoint is reliable without authentication. Ready to merge.

kaifcodec (approved):
@anonymort Thanks for the update, this approach is more reliable as it directly checks the username availability using the account creation flow.

PR #332, opened 2026-04-30
beatstars user check

notlorick:


kaifcodec:
@notlorick Thanks for opening the PR, I’ve reviewed both modules.

`beatstars.py` looks good and works as expected. However, the VK module is returning a `301` status code. Could you test it again on your side and confirm?

notlorick:
code is a bit uglier now, but it should work as expected

kaifcodec:
@json-hunter07 Take a look at this. If you think any improvements can be made to the error handling, feel free to proceed with them.

json-hunter07:
@kaifcodec I have improved the error handling and edge cases in the `beatstars` module.
Now it correctly handles the error.
However I also do think that instead of reporting the `invalid length` as an error to reporting it as not found but passing the message would be a good idea.

kaifcodec:
@json-hunter07 Good improvement on the error handling and edge cases in the `beatstars` module.

And yes, I also think treating `invalid length` as `not found` while still passing the message/details would make more sense than reporting it as a hard error.

kaifcodec (approved):

json-hunter07 (approved):

kaifcodec:
@notlorick Thank you for the PR, if want to contribute further feel free to add the email validation of these sites or any new sites that aren't already covered.

PR #333, opened 2026-05-13
refactor: support multiline extra fields and add Etsy OSINT module

kaifcodec:
- Update Result class console output to dynamically parse and render multiline 'extra' strings using tree-node formatting (`└── `).
- Fix `to_csv` parsing to flatten multiline string outputs into inline semicolon-separated rows, preventing layout breaking.
- Fix `to_json` string formatting to safely escape newline characters and internal quotes inside the hardcoded JSON template.
- Implement `validate_etsy` module to scrape comprehensive profile metrics (ID, privacy flags, shop/seller status, timestamps, and high-res avatars).

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
   <img src="https://img.shields.io/pypi/dm/user-scanner?style=for-the-badge" />
 </p>
 
+
 ---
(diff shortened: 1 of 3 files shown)


json-hunter07 (approved):
@kaifcodec LGTM,
The modifications in `Result` class is great for future module additions.

PR #335, opened 2026-05-16
Fix: missing csv "extra" header

VamatoHD:
The "extra" header in CSV files was missing. This PR enables the  header to automatically update based on the `CSV_TEMPLATE` in `result.py`.


kaifcodec (approved):
Good catch @VamatoHD !

PR #338, opened 2026-05-16
fix(gumroad, girlslife): force Accept-Encoding: identity

sturq:
Both modules advertised `Accept-Encoding: gzip, deflate, br, zstd`, but httpx ships only gzip/deflate decoders by default. When servers honored the header and replied with brotli or zstd, the response body came back as raw compressed bytes, breaking the nonce regex in girlslife and `response.json()` in gumroad:

* `Girlslife: Failed to parse response`
* `Gumroad: unexpected exception:`

Switching both to `Accept-Encoding: identity` keeps the response uncompressed, no extra deps required.

Verified locally with brotli/zstandard NOT installed (httpx decoders limited to `identity, gzip, deflate`): both modules return `Registered` / `Not Registered` cleanly.

Per your comment on the previous revision, scoped this to only the two confirmed-broken modules. Same pattern exists in a dozen other places (orchestrator, both `github.py`, foxnews, ama, dribbble, linktree, producthunt, launchpad, battlenet, snapchat, threads), but those are currently working since their upstream servers are falling back to gzip. Happy to follow up with a separate PR if any of those start failing.

### DIFF
--- a/user_scanner/email_scan/creator/gumroad.py
+++ b/user_scanner/email_scan/creator/gumroad.py
@@ -11,7 +11,7 @@ async def _check(email: str) -> Result:
             headers1 = {
                 'User-Agent': "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36",
                 'Accept': "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8",
-                'Accept-Encoding': "gzip, deflate, br, zstd",
+                'Accept-Encoding': "identity",
                 'sec-ch-ua': '"Not(A:Brand";v="8", "Chromium";v="144", "Google Chrome";v="144"',
                 'sec-ch-ua-mobile': "?0",
                 'sec-ch-ua-platform': '"Linux"',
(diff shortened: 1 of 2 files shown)


kaifcodec:
@sturq Thank you for the heads up, you’re correct about this.

Some modules can indeed break on different environments because of the advertised `Accept-Encoding` values. However, instead of introducing new dependencies for Brotli and Zstandard support, we can likely fix most of these cases by forcing `Accept-Encoding: identity` on the affected modules.

You can try this yourself: if a module is failing because of compressed responses, replace the current `Accept-Encoding` value with just `identity` and test it again. If it resolves the issue, feel free to open a PR for those specific modules.

In general, using `identity` is the safest option. Some modules were left with default/browser-like encodings simply because they worked at the time, but it’s possible that certain servers have since changed their behavior and are now returning `br` or `zstd` responses, which would require explicitly forcing `identity`.

sturq:
Makes sense, agree on avoiding new deps. I've reset the branch and force-pushed a new commit that just switches gumroad and girlslife to `Accept-Encoding: identity` instead. Both confirmed working locally with brotli/zstandard not installed (httpx falls back to identity/gzip/deflate only) and the modules return real results. Scoped this PR to those two since they're the ones throwing visible errors. Left the rest alone for now since their servers seem to be falling back to gzip, can follow up if any of them start regressing.

kaifcodec (approved):
@sturq This is the correct approach, It keeps us safe from future changes.

PR #352, opened 2026-05-30
fix(orchestrator): enable TLS certificate verification in HTTP client

andesyteoss:
## Summary

The shared HTTP client factory `get_client()` in `user_scanner/core/orchestrator.py` explicitly disables TLS certificate verification by passing `verify=False` to `httpx.Client`. Since `get_client()` is the centralized client used by `make_request()`, `generic_validate()`, and `status_validate()`, this affects every outbound HTTPS request made by the scanner — over 300 call sites across all user-scan and email-scan modules.

With TLS verification disabled, an attacker in a man-in-the-middle position (e.g., on the same Wi-Fi network, a compromised router, or a malicious proxy) can intercept and read all traffic between the scanner and the target services. This traffic includes the usernames and email addresses being scanned — which is the core sensitive data the tool handles.

**CWE-295: Improper Certificate Validation**

## The Fix

This is a one-line change on line 145 of `user_scanner/core/orchestrator.py`:

```diff
- _clients[key] = httpx.Client(http2=use_http2, proxy=proxy_val, verify=False)
+ _clients[key] = httpx.Client(http2=use_http2, proxy=proxy_val, verify=True)
```

`verify=True` is actually httpx's default, so this change restores standard behavior. The `verify=False` was likely added during development to work around certificate issues and never reverted.

## Why this matters

The tool's purpose is to check whether a username or email is registered on various platforms. Users provide their own identifiers (or targets') as input. With `verify=False`:

1. All scanned usernames/emails are transmitted over connections that don't validate the server's identity
2. An attacker on the network path can silently intercept this data
3. The attacker can also inject modified responses, potentially causing the scanner to report incorrect availability results

## Proof of Concept

An attacker on the same network can use mitmproxy to intercept all scanner traffic:

```bash
# Terminal 1: Start mitmproxy on port 8080
mitmproxy --mode regular --listen-port 8080

# Terminal 2: Run the scanner through the proxy
# With verify=False (current), this succeeds — mitmproxy sees all usernames in cleartext
HTTPS_PROXY=http://127.0.0.1:8080 python -m user_scanner scan testuser123

# After this fix (verify=True), the same command fails with a TLS certificate error,
# because mitmproxy's certificate is not trusted — protecting the user's data.
```

Without the fix, mitmproxy displays every request URL containing the scanned username. With the fix, the connection is refused because the proxy's self-signed certificate fails validation.

## Testing

- Verified the fix compiles and the module loads without errors
- Confirmed `verify=True` is httpx's default behavior, so no functionality regression is expected for users on networks with valid certificates
- Checked that no other `httpx.Client` instantiations in the codebase use `verify=False` (the one in `helpers.py:132` and `hudson.py:67` both use defaults, which is `verify=True`)
- The change is minimal (1 line) and isolated to the client factory

## Adversarial review

Before submitting, we considered whether this is actually exploitable in practice. The main question was whether the MITM prerequisite makes this purely theoretical. It doesn't — users commonly run OSINT tools on shared/public networks (coffee shops, conferences, coworking spaces), which is exactly where MITM attacks are most feasible. Additionally, the tool supports proxy configuration (`-p` flag), and a malicious proxy would trivially exploit `verify=False`. The fix has zero downside for legitimate use and closes a real exposure window.

### DIFF
--- a/user_scanner/core/orchestrator.py
+++ b/user_scanner/core/orchestrator.py
@@ -142,7 +142,7 @@ def get_client(use_http2: bool, proxy_val: Optional[str]) -> httpx.Client:
     if key not in _clients:
         with _clients_lock:
             if key not in _clients:
-                _clients[key] = httpx.Client(http2=use_http2, proxy=proxy_val, verify=False)
+                _clients[key] = httpx.Client(http2=use_http2, proxy=proxy_val, verify=True)
     return _clients[key]
 
 


kaifcodec (approved):
@sebastiondev Thanks for the PR, this is a reasonable fix!

andesyteoss:
Thanks for the merge. Glad it landed cleanly.

PR #355, opened 2026-05-31
refactor: enable deep username scan for user_scan modules

kaifcodec:
* **Dynamic OSINT Metadata Extraction**: Upgraded 20 Social, Developer, and Music check modules to extract public profile metadata (bios, locations, follower/following counts, join dates, and avatar URLs) dynamically.
* **Defensive Parsing Wrappers**: Wrapped all JSON, GraphQL, and HTML regex parsers in try-except scopes to ensure graceful degradation to basic taken status if site structures change.
* **API Rate-Limit Bypasses**: Migrated modules (like GitHub) from signup APIs to public profile scraping to prevent unauthenticated API rate limits.
* **Upgraded Social Modules (`user_scan/social/`)**:
  * `anilist`, `bluesky`, `mastodon`, `openstreetmap`: Scrapes GraphQL, NEXT_DATA, and HTML to pull detailed bio, location, followers count, and avatars.
  * `35photo`, `pinterest`, `snapchat`: Extracts photographer stats, redux state JSON, and snapcode SVG profiles.
* **Upgraded Developer Modules (`user_scan/dev/`)**:
  * `github`: Direct profile scraping of names, bios, location, public emails, and social links using schemas (`itemprop="image"`) to avoid sponsor avatar conflicts.
  * `gitlab`, `huggingface`, `dockerhub`, `cratesio`: Switched to public APIs to capture user IDs, joined dates, active states, and avatar URLs.
* **Upgraded Music Modules (`user_scan/music/`)**:
  * `soundcloud`, `bandcamp`: Extracts embedded hydration states and data-blobs containing full user details, followers, and play statistics.
  * `discogs`, `bandlab`, `freesound`, `lastfm`, `audiojungle`, `beatstars`: Scrapes AJAX stats, public REST APIs, and profile headers to extract comprehensive artist registries and music metrics.

---
  
  TODO:
  

- [x]  Refactor and migrate rest of the modules

- [x]  Test all the modules if any data extraction left behind or got ignored

- [x] Update docs

PR #357, opened 2026-06-12
added pypi

syswraith:
nothing more needed than a status_validate :)

### DIFF
--- a/user_scanner/user_scan/dev/pypi.py
+++ b/user_scanner/user_scan/dev/pypi.py
@@ -1,17 +1,144 @@
 import re
+import xmlrpc.client
+import httpx
 
-from user_scanner.core.orchestrator import Result, status_validate
+from user_scanner.core.helpers import get_random_user_agent
+from user_scanner.core.orchestrator import Result, make_request
 
 
-def validate_pypi(user):
+def validate_pypi(user: str) -> Result:
+    """
+    Validates a PyPI username and extracts:
+
+    - display_name
+    - email
+    - packages_count
+    - packages
+    """
+
     if not re.match(r"^(?!_+$)[A-Za-z0-9._-]+$", user):
         return Result.error(
             "Username may only contain letters, numbers, periods, underscores, and hyphens, and cannot consist solely of underscores"
         )
 
-    url = f"https://pypi.org/user/{user}"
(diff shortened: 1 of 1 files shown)


syswraith:
by the way, i have a question
can i implement the logic for multiple sites/services and then open 1 pull request for all of them or do i have to open it for each one?

kaifcodec:
@syswraith Thank you for you adding this, however we just moved from normal status check to deep scan mode yesterday check the PR #355 . If pypi expose any type of user data in the reponse you can add those too and edit this module, you can check the new `../user_scan/dev/github.py` module  to understand how we handle the `extra` field.

And for your question, yes you can add multiple sites in a single commit and open a single PR.

kaifcodec:
@syswraith  I just tested it myself and I noticed it returns `found` for all the username, there is a false positive issue.
Check the module yourself one more time, however `pypi` is very aggressive in bot detection.

syswraith:
yes i noticed this as well, im working on a fix + getting pfp, date of joining and name

kaifcodec:
PyPI's web user profiles (`https://pypi.org/user/{username}`) are protected behind a Fastly JavaScript challenge. However, the XML-RPC API at `https://pypi.org/pypi` remains open. you can use it to fetch a user's packages and determine if the username is taken.

You have to use this `import xmlrpc.client` and POST request.
...
xml_payload = xmlrpc.client.dumps((user,), methodname="user_packages")


If you take my route.

syswraith:
since fastly blocks the /user page, this is a nice workaround
sadly this also means that we cant get the date of joining or the pfp
(gravatar pfp) but thats alright i think
might get the email if its present tho so thats nice
and the packages

kaifcodec:
It's perfect, but there could be an issue i.e. if a user has too many packages, then that list can make the console output unnecessarily bloated. So it would be better if you set a limit to extract only around 15 or 20 packages and append it to the list.

syswraith:
TIL package names can be ridiculously long, which is messing up the output. Capped the limit at 5 instead. Tested of the user `aws`.

kaifcodec (approved):
@syswraith THANK YOU! FOR THIS ADDITION.
If you want to contribute further, you are always welcome.

PR #361, opened 2026-06-13
added luarocks!

syswraith:
hello there

Works on my machine `¯\_(ツ)_/¯`

### DIFF
--- a/user_scanner/user_scan/dev/luarocks.py
+++ b/user_scanner/user_scan/dev/luarocks.py
@@ -157,7 +157,12 @@ def validate_luarocks(user: str) -> Result:
         re.IGNORECASE | re.DOTALL,
     )
 
-    extra["modules_list"] = [name.strip() for _, name, _ in modules[:3]] + ["..."]
+    if not modules:
+        extra["modules_list"] = None
+    else:
+        extra["modules_list"] = [name.strip() for _, name, _ in modules[:3]]
+        if len(modules) > 3:
+            extra["modules_list"].append("...")
 
     return Result.taken(
         url=url,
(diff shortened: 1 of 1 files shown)


user_scanner/user_scan/dev/luarocks.py
```diff
@@ -0,0 +1,165 @@
...
+    - twitter
+    - website
+    - registered
+    - modules_count
+    - downloads
+    - modules_list (first 3)
+    """
+
+    if not re.match(r"^[A-Za-z0-9._-]+$", user):
+        return Result.error(
```
kaifcodec:
```suggestion
        return Result.available(
```

This one will be better with our newer approach of username OSINT.

kaifcodec (changes requested):

kaifcodec:
Also around line 135 you can add a module counter first and then later use it before you show `modules_list`, to check if the modules count is only three, greater or lower. Then it should skip printing `[...]` for `<=3` and also an `if/else` block like below,

```python
    if not modules:
        extra["modules_list"] = None
    else:
        # here the if/else to check the modules count and decide if to print the ["..."]
        extra["modules_list"] = [name.strip() for _, name, _ in modules[:3]] + ["..."]
```
It makes it logically correct.

syswraith:
check now

kaifcodec (approved):
@syswraith It's perfect now. I will take look on this and test it myself after a while and merge it.

Thank you again!

Till then can you update the old `CONTRIBUTING.md`? As we moved from simple username availability checker to deep username scans which extract data. (Just the user_scan mode guide needs the change)

PR #378, opened 2026-06-17
fix(email_scan): retire dead boot.dev email module

### LINKED ISSUE #339: Boot.dev module: upstream removed /v1/users/email/exists (HTTP 404)
The `boot_dev` module hits `POST https://api.boot.dev/v1/users/email/exists`, which now returns 404:

```
$ curl -s -X POST "https://api.boot.dev/v1/users/email/exists" \
    -H "Content-Type: application/json" \
    --data '{"email":"test@example.com"}' -w "\nHTTP %{http_code}\n"
404 page not found
HTTP 404
```

Looking at the production Nuxt bundles on `www.boot.dev`, there's no email-existence endpoint left under `/v1/` at all. The only email-related paths are `/v1/users/email/reset`, `/v1/users/email/send_reset_link`, and `POST /v1/users` (which actually creates a user and requires first/last name). So there's no unauthenticated, side-effect-free way to probe existence anymore.

Two options:

1. Remove the module.
2. Reimplement it as a loud check against `/v1/users/email/send_reset_link` (would need to confirm the response distinguishes exists vs not), and add it to the loud-skip list.

I can put up a PR for either if you tell me which way you want to go.

Tested on user-scanner 1.3.6.4, Python 3.13.13, Termux.

Brumbelow:
boot.dev removed the unauthenticated email-existence endpoint — `POST https://api.boot.dev/v1/users/email/exists` now 404s:
```
    curl -s -X POST "https://api.boot.dev/v1/users/email/exists" \
      -H "Content-Type: application/json" \
      --data '{"email":"test@example.com"}' -w "\nHTTP %{http_code}\n"
```
There's no unauthenticated way to probe existence anymore as far as I can tell. Moved it to `abandoned/email_scan/` to match how codepen/saashub were retired, rather than
deleting it outright. The username checker (`user_scan/dev/boot_dev.py`) scrapes the profile page, not the API, so it's left untouched.

Closes #339


kaifcodec:
@Brumbelow Thank you for the heads up.
Can you create another folder inside `email_scan/` i.e. `dev/` and move all 3 of the modules? Also in  `user_scan/` create a new folder `adult/` and move adult_forum.py to it.

Brumbelow:
@kaifcodec created and moved

kaifcodec (approved):
Thank you @Brumbelow 
Will be glad to see you with more contributions.

PR #379, opened 2026-06-19
add carrd and strawpage support

ghost:
tested on invalid and valid usernames

### DIFF
--- a/user_scanner/user_scan/other/carrd.py
+++ b/user_scanner/user_scan/other/carrd.py
@@ -1,0 +1,23 @@
+from user_scanner.core.helpers import get_random_user_agent
+from user_scanner.core.orchestrator import Result, make_request
+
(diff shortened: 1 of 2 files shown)


ghost:
its showing a different commit author than me for some reason i dont know if that will effect anything

kaifcodec:
@fudness Thanks for opening the PR.
I know that our CONTRIBUTING.md isn't updated yet for the new feature we have added, that is we moved from simple status check (`Found` or `Not Found` check to deep data extraction) you can see the examples in `../user_scan/dev/github.py`.
Can you test the modules again if they return some data of the user which can be or will be good to extract?

And for the second question yes it will cause some issues if the commit author is different from PR author. Use the below commands on your terminal and set the email of your current github account,
```bash
git config --global user.name "Your Name"
git config --global user.email "your_email@example.com"

```

PR #360, opened 2026-06-13
Added a basic nix flake

BoxC4T:
adds a simple nix flake to allow nix based users to quickly use user-scanner
unless the project majorly changes the only thing that would require updates is just the version string on line 18

PR #384, opened 2026-06-21
added automatic version pulling from py toml to the nix flake

BoxC4T:
found the function to pull from a toml. This should mean the only times that changes to the flake are required would be a change in file struct or dependancys

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,6 +7,7 @@
     self,
     nixpkgs,
   }: let
+    pyproject = fromTOML (builtins.readFile ./pyproject.toml);
     systems = [
       "x86_64-linux"
       "aarch64-linux"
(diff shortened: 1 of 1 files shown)


kaifcodec:
@BoxC4T Thank you for opening the PR.

I already had this feature in mind. I’ve just been holding off because I want (or need) to keep updating the file manually for a little while longer before automating it. I’ll most likely merge this after the next version bump (`v1.4.0.3`).

One thing I noticed, both in this PR and PR #360, is that your commit messages aren’t using conventional commit types (e.g. `feat`, `fix`, `refactor`, etc.).

I’d recommend reopening this PR with a properly typed commit message. For example:

- `feat: add automatic version pulling from pyproject.toml to flake.nix`
- `refactor(flake.nix): automatically pull version from pyproject.toml`

That way, we can merge it directly from your branch (without needing to squash it) and keep the commit history consistent.

PR #385, opened 2026-06-22
fix(dribbble): prevent regex overflow on empty profiles

Priyanshu-x:
**Summary**

This PR fixes a parsing bug in the Dribbble module that could cause large portions of raw HTML to be printed when scanning profiles without a custom bio. It also improves profile verification to reduce false positives from non-profile pages that return HTTP 200 responses.

**Root Cause**

Dribbble parser used a regex that could match beyond the intended attribute boundary:
([^\|]+?)
to extract profile metadata.

For profiles whose meta description did not contain a pipe (|), the regex could continue matching beyond the intended attribute boundary and consume large portions of the HTML document. The captured HTML was then rendered as profile metadata in the terminal output.

**Fix**

Regex Confinement
Bound extraction to the content="..." attribute.
Parse the extracted content using Python string operations rather than multiple regexes.
Prevent HTML from being captured as metadata.
Empty Profile Handling
Ignore Dribbble's default site-wide description (Find Top Designers...).
Gracefully handle profiles that do not provide custom metadata.
Profile Verification
Extract and compare canonical/og values instead of relying solely on HTTP 200 responses.
Normalize URLs using lowercase comparison and trailing-slash removal.
Improve handling of redirects and soft-404 scenarios.

**Regression Testing**

Verified against:

✅ dribbble (standard profile)
✅ Rahul (empty-profile case that triggered the bug)
✅ google (non-standard profile structure)
✅ microsoft (standard profile)
✅ johndoe (missing profile)

Comparision-user_scanner

### DIFF
--- a/user_scanner/user_scan/creator/dribbble.py
+++ b/user_scanner/user_scan/creator/dribbble.py
@@ -17,20 +17,40 @@ def validate_dribbble(user):
         response = make_request(url, headers=headers, follow_redirects=True)
         if response.status_code == 200:
             html = response.text
-            extra = {}
-            name_match = re.search(r'<meta name=\"description\" content=\"([^\|]+?)\s*\|', html)
-            if name_match:
-                extra['name'] = name_match.group(1).strip()
+            
+            # Explicit verification: Dribbble normalizes casing and strips trailing slashes in its tags
+            # We extract canonical or og:url and compare case-insensitively
+            canonical_match = re.search(r'<link rel="canonical" href="([^"]+)"', html, re.I)
+            og_match = re.search(r'<meta property="og:url" content="([^"]+)"', html, re.I)
+            
+            found_url = ""
+            if canonical_match:
+                found_url = canonical_match.group(1)
+            elif og_match:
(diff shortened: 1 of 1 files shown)


kaifcodec:
@Priyanshu-x  Thank you for opening this PR and I am really impressed by your clean PR description.

Your logic is correct and it really fix the bug, however I just noticed I can make the data extraction better because right now it doesn't really extract every data and dribble has a dedicated API endpoint `/about` which exposes most of the important data about the user.

I will push that commit shortly and merge it.

Priyanshu-x:
@kaifcodec Thanks for reviewing it and for the kind words. Using the dedicated /about endpoint sounds like a much cleaner approach. Glad the investigation and root cause analysis were helpful. Looking forward to the update and merge.

kaifcodec (approved):
I am done with my reverse engineering and now I am going to merge it.

However if you want to contribute further, you are always welcome.

json-hunter07 (approved):
LGTM @kaifcodec

PR #396, opened 2026-07-01
feat: add Zillow module

shredzwho:
This PR adds a new shopping module for Zillow.

### DIFF
--- a/tests/test_zillow_module.py
+++ b/tests/test_zillow_module.py
@@ -1,0 +1,25 @@
+import sys
+from pathlib import Path
+
+# Add project root to sys.path
+sys.path.append(str(Path(__file__).resolve().parent.parent))
+
+from user_scanner.user_scan.shopping.zillow import validate_zillow
+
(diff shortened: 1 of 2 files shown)


kaifcodec:
@shredzwho We don't need a separate `test` file for each module, so you can remove it.

One thing we've recently introduced is the `extra` parameter in `Result.taken()` to pass any additional data that a site exposes about a user. If this site returns profile metadata publicly, try extracting those details with regex (or another appropriate parser) and pass them through `extra`.

If the profile data is only available after login or signup, let me know and I will merge it as it is for now.

Azaucifer:
Hey @shredzwho! 👋

Just checking in on this PR. It's been open for a while and I noticed it has the `help wanted` label.

I'd be happy to help finish it up by:

- Removing the test file (as requested)
- Adding extra data extraction for any public profile metadata

Are you still working on this? If not, I can pick it up and open a new PR with the changes.

Let me know! 🙌

PR #403, opened 2026-07-04
fix(duolingo): extract additional user data like streak and picture

### LINKED ISSUE #401: Enrichment Modules
I think that a feature where you got data from an email would be nice, I don't just mean if a email is registered or not on a website but instead give more info about the account tied to the email in the website, for example through the duolingo api you can lookup an email(https://www.duolingo[.]com/2017-06-30/users?email=examplegmail.com) and it returns an ID, then on another endpoint from duolingo you can further elaborate from that ID all the way to full account info. I think that's something taking a look at because it could definitely help with investigations.

Thanks.

kaifcodec:
Closes #401

### DIFF
--- a/user_scanner/email_scan/learning/duolingo.py
+++ b/user_scanner/email_scan/learning/duolingo.py
@@ -23,7 +23,29 @@ async def _check(email: str) -> Result:
                 data = response.json()
                 # Duolingo returns a list of users matching the email
                 if data.get("users") and len(data["users"]) > 0:
-                    return Result.taken(url=show_url)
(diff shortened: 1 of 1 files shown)


json-hunter07 (approved):
LGTM,
btw thanks to @detxctive !

kaifcodec:
Yeah we missed that part.

detxctive:
Thanks, I'm also pretty sure it used to show way more info: https://web.archive.org/web/20251005124909/https://blackowlintelligence.com/osint/4

PR #406, opened 2026-07-06
feat: add user modules for Yaga, StackB, Pedsovet

kristoisberg:
This PR adds username modules for several sites I've ran into with usernames exposed in their profile URLs:

- Yaga is an Estonian online marketplace, they operate sites for Estonia, Latvia and South Africa, but the Latvian shares its database with the Estonian one so I omitted that one.
- StackB is a Russian gaming social network.
- Pedsovet is an ancient Russian site for study materials.

### DIFF
--- a/tests/test_pedsovet.py
+++ b/tests/test_pedsovet.py
@@ -1,0 +1,111 @@
+from types import SimpleNamespace
+
+from user_scanner.core.result import Status
+from user_scanner.user_scan.other import pedsovet
+
+
+def _profile_html(username="asd"):
+    return f"""
+    <html>
+      <body>
+        <div class="user">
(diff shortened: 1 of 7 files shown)


kaifcodec:
@kristoisberg Thank you for opening the PR.
However this PR have some issues.
1. We don't create individual tests for each sites, so you should remove them.
2. The validation modules are unnecessarily big, due to creation of helper functions inside the modules and most of them don't align with existing module structure. I suggest you to go through the `CONTRIBUTING.md` or take a look into existing modules like `github.py`, `instagram.py` etc. in `user_scan` directory to understand how the `extra` are passed and `validate_` functions are created. **After that open a new PR with only the validation modules.**

PR #409, opened 2026-07-07
feat: add user modules for Yaga, StackB, Pedsovet

kristoisberg:
Seriously sorry for the previous pull request. I removed the per-module tests and parser fallbacks, which made the modules too complex (they are still more complex than most, since these are not the most simple sites to parse). I also made the two Yaga modules completely self-contained and refactored the modules.

Original description:

> This PR adds username modules for several sites I've ran into with usernames exposed in their profile URLs:
> 
> - Yaga is an Estonian online marketplace, they operate sites for Estonia, Latvia and South Africa, but the Latvian shares its - database with the Estonian one so I omitted that one.
> - StackB is a Russian gaming social network.
> - Pedsovet is an ancient Russian site for study materials.


kaifcodec:
@kristoisberg Thank for the PR, I am tied up with some work so I was unable to test these modules manually, whenever I become free I will test them and let you know.

kaifcodec:
@kristoisberg First off, thank you very much for your patience! 

I tested the 4 modules you added. Pedsovet and Yaga.ee are both working fine and handling extraction perfectly. However, Stackb is returning `Not Found` for common usernames like `johndoe` and `alex` (looks like a false negative issue). As for Yaga.co.za, it's throwing a 403 error on every scan, which is likely due to bot detection. 

If these two modules are fixable, you can go ahead and patch them. Otherwise, feel free to just remove them from the PR. 

Let me know how you want to proceed with this!

kristoisberg:
Hey @kaifcodec, thanks for the review!

I ran some tests of my own and found the following:
-  yaga.co.za appears to be region blocked in a large part of the world as shown here: https://check-host.net/check-http?host=https://www.yaga.co.za/
- StackB legitimately seems to be missing the common usernames you tested with. I did find an issue causing some false negatives - they treat usernames case-sensitively in the profile URL. However, I found an alternative way of checking that is case-insensitive and should also allow for an email module for StackB. Enrichments would only work for usernames with a case-sensitive match though...

kaifcodec:
@kristoisberg Thank you for the detailed information about it.
We will keep the yaga.co.za module and change the error message to the correct one, like, if it throws 403 that means either forbidden due to region block or bot detection/captcha.

However, do you want to proceed with the alternative method you suggested here for StackB or remove the module completely?

kristoisberg:
I updated the StackB module to use the login form instead. I also slightly updated Yaga modules to check for a 404 response code - I don't recall the site responding with that code before, but now it was.

kaifcodec (approved):
@kristoisberg I tested all the modules locally, what I noticed is that stackb module still throwing errors like,
- `Unexpected login check response`
- `ConnectTimeout: _ssl.c:1063: the handshake operation timed out`

even when i tried it in 4 to 5 different region proxies.

So I removed `stackb.py` module from this PR branch as other modules are working and I don't want this PR to hang for long so I'm merging this PR to `main` without stackb.py, if you find the fix for it later, feel free to open a new PR with the fixed module.

**Thank you for the contribution!**

PR #417, opened 2026-07-18
feat: add RedTube email scan module

brunolm:
## What

Adds an email-based **RedTube** account check to `email_scan/adult/`, filling one of the missing adult modules.

## How it works

Mirrors the existing `pornhub.py` pattern (RedTube is the same MindGeek/Aylo network):

1. `GET /register` to scrape the per-session `page_params.token` (note: the token ends in a literal `.` that must be preserved, or it's rejected).
2. `POST /user/create_account_check` with `check_what=email` — RedTube returns the same structured JSON as Pornhub.

Response mapping:

| Response | Result |
|----------|--------|
| `create_account_passed` | Not Registered |
| `create_account_failed` + `already registered` | Registered |
| `create_account_failed` + "does not meet our registration requirements" (valid email) | Registered |
| bare `"NOK"` string / HTTP 429 | Rate limited (error) |

## Shared email validation

Extracted a reusable `is_valid_email` helper into `core/helpers.py` (previously an inline regex duplicated twice in `__main__.py`). The new regex is stricter and more reliable, inspired by the `email-validator` package:

- proper dot-atom local part (rejects leading/trailing/double dots)
- valid domain labels (rejects hyphen-edge and empty labels)
- alphabetic TLD (≥2 chars; rejects 1-char and numeric TLDs)
- RFC 5321 length limits and `fullmatch` (rejects trailing-newline edge case)

The two existing CLI call sites now reuse the helper.

## Testing

- Verified live against the RedTube endpoint (passed / registered / rate-limited paths).
- Validator checked against the top 10 mainstream email providers and a range of valid/invalid edge cases.
- `pytest tests/test_helpers.py`: all email-format tests pass (the 2 `*_unreadable` failures are pre-existing Windows `chmod(0)` no-ops, unrelated to this change).

### DIFF
--- a/user_scanner/__main__.py
+++ b/user_scanner/__main__.py
@@ -1,7 +1,6 @@
 import argparse
 import json
 import os
-import re
 import sys
 import time
 
@@ -22,6 +21,7 @@ from user_scanner.core.helpers import (
     get_proxy_count,
     get_site_name,
     is_loud,
+    is_valid_email,
     load_categories,
     load_modules,
     set_proxy_manager,
@@ -233,7 +233,7 @@ def main():
             # Validate email formats
             valid_emails = []
             for email in emails:
-                if re.findall(r"^[^@\s]+@[^@\s]+\.[^@\s]+$", email):
+                if is_valid_email(email):
                     valid_emails.append(email)
                 else:
                     print(f"{Y}[!] Skipping invalid email format: {email}{X}")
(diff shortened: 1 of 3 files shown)


kaifcodec (approved):

kaifcodec:
@brunolm Thank you for the PR and the helper function for email validation.

PR #418, opened 2026-07-18
feat: add xHamster username scan module

brunolm:
## What

Adds a username-based **xHamster** profile check to `user_scan/adult/`, filling one of the missing adult modules.

## How it works

Follows the existing `user_scan/adult` pattern (synchronous `validate_xhamster(user)` via `generic_validate`). It requests the current canonical profile path `https://xhamster.com/users/profiles/{user}`:

| Response | Result |
|----------|--------|
| 200 + `<link rel="canonical">` echoing `/users/profiles/{user}` | Found |
| 404 | Not Found |
| 200 without a matching canonical (age gate / landing) | Error: "Profile confirmation not found" |
| other status (e.g. 301 reserved names) | Error: "Unexpected status: {code}" |

The canonical-link confirmation (rather than status-200 alone) guards against age-gate/landing pages being misread as hits. Canonical extraction isolates the `<link rel="canonical">` tag before pulling `href`, so it is independent of HTML attribute order.

## Notes

- xHamster exposes no email-existence endpoint, so this is a **username** module (`user_scan`), not an email one — consistent with how WhatsMyName/Sherlock treat the site.
- Uses the current `/users/profiles/{user}` path directly; the legacy `/users/{user}` path now 301-redirects to it.

## Testing

- Verified live: existing usernames (`john`, `admin`, `test`, `Mike`) → Found; non-existent → Not Found; case-insensitive match confirmed.

### DIFF
--- a/user_scanner/user_scan/adult/xhamster.py
+++ b/user_scanner/user_scan/adult/xhamster.py
@@ -1,0 +1,37 @@
+import re
+from user_scanner.core.orchestrator import generic_validate, Result
+
+
+def validate_xhamster(user):
+    url = f"https://xhamster.com/users/profiles/{user}"
+    show_url = url
+
+    def process(response):
+        if response.status_code == 404:
+            return Result.available(url=show_url)
+
+        if response.status_code != 200:
+            return Result.error(f"Unexpected status: {response.status_code}", url=show_url)
+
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@brunolm That's some excellent work!
I mean, I am genuinely impressed by seeing the amount of data it extracts from the HTML response!

PR #420, opened 2026-07-18
feat: add YouPorn email scan module

brunolm:
## What

Adds an email-based **YouPorn** account check to `email_scan/adult/`.

## How it works

YouPorn is on the same Aylo/MindGeek backend as Pornhub/RedTube. It uses the signup form's async validation endpoint:

1. `GET /register` to scrape the per-session `page_params.token` (ends in a literal `.` that must be preserved).
2. `POST /register/verify_email` with `token` + `email`.

Response mapping:

| Response | Result |
|----------|--------|
| `{"success":true}` | Not Registered |
| `{"success":false, ["...does not meet our registration requirements."]}` (valid email) | Registered |
| `{"success":false, ["Not available."]}` (stale/over-used token throttle) | Error (rate limited) |
| non-200 / 429 | Error |

For a well-formed address, YouPorn reports an existing account by refusing registration ("does not meet our registration requirements") rather than saying it's taken — same behavior as RedTube. The `is_valid_email` guard ensures malformed inputs aren't misreported as hits.

## Shared helper

Reuses the `is_valid_email` helper in `core/helpers.py`. This is the **same helper introduced alongside the RedTube module** (PR #417) — byte-for-byte identical — so if both land, the second merge is a no-op on `helpers.py` rather than a real conflict.

## Notes

- **Email-only:** YouPorn user profiles are addressed by numeric ID (`/users/0000/0286/…`), so username lookups by name 404 — no viable username module.
- Neither holehe nor MailAccess has a YouPorn checker (only a stale Sherlock username entry pointing at the now-404 `/users/{name}` path).

## Testing

- Verified live: `brunolm7@gmail.com` → Registered; throwaway addresses → Not Registered.

### DIFF
--- a/user_scanner/core/helpers.py
+++ b/user_scanner/core/helpers.py
@@ -7,6 +7,7 @@ import inspect
 import json
 import os
 import random
+import re
 import threading
 import functools
 import asyncio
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
@brunolm Outstanding work on these modules. The quality of your recent PRs has been excellent and brings great value to the project.

If you have the bandwidth, we would love to see your continued contributions. We are also very open to inviting you to join us as a core maintainer if you have the time to dedicate to a larger role. Let us know your thoughts!

kaifcodec:
I have personally tested the modules and everything looks great; I will test a few edge cases and get them merged soon!

PR #421, opened 2026-07-18
feat: add Tube8 email scan module

brunolm:
## What

Adds an email-based **Tube8** account check to `email_scan/adult/`.

## How it works

Tube8 is on the same Aylo/MindGeek backend as Pornhub / RedTube / YouPorn and uses the identical signup validation flow:

1. `GET /register` to scrape the per-session `page_params.token` (ends in a literal `.` that must be preserved).
2. `POST /register/verify_email` with `token` + `email`.

Response mapping:

| Response | Result |
|----------|--------|
| `{"success":true}` | Not Registered |
| `{"success":false, ["...does not meet our registration requirements."]}` (valid email) | Registered |
| `{"success":false, ["Not available."]}` (stale/over-used token throttle) | Error (rate limited) |
| non-200 / 429 | Error |

For a well-formed address, Tube8 reports an existing account by refusing registration ("does not meet our registration requirements") rather than saying it's taken — same behavior as YouPorn/RedTube. The `is_valid_email` guard ensures malformed inputs aren't misreported as hits.

## Shared helper

Reuses the `is_valid_email` helper in `core/helpers.py`. This is the **same helper introduced alongside RedTube** (PR #417) and YouPorn (PR #420) — byte-for-byte identical — so overlapping merges resolve as a no-op on `helpers.py`.

## Notes

- **Email-only:** Tube8 user profiles are addressed by numeric ID (Aylo platform), so username lookups by name aren't viable.
- Neither holehe nor MailAccess has a Tube8 checker.
- The module is a domain-swap of the YouPorn module.

## Testing

- Verified live: `test@gmail.com` → Registered; throwaway addresses → Not Registered.

> Note: like the other Aylo sites, the endpoint's anti-enumeration throttle can intermittently return the "requirements" message for a valid address, so a Registered result on a valid email is high-signal but not a hard guarantee; Not Registered is reliable.

### DIFF
--- a/user_scanner/core/helpers.py
+++ b/user_scanner/core/helpers.py
@@ -7,6 +7,7 @@ import inspect
 import json
 import os
 import random
+import re
 import threading
 import functools
 import asyncio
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
LGTM.

PR #422, opened 2026-07-19
feat: add SpankBang username scan module

brunolm:
## What

Adds a username-based **SpankBang** profile check to `user_scan/adult/`, routed through a browser-impersonating request so it clears Cloudflare's TLS-fingerprint bot wall.

## How it works

Requests `https://spankbang.com/profile/{user}`:

| Response | Result |
| --- | --- |
| 200 + title `"{user} Profile … @ SpankBang"` | Found |
| 404 | Not Found |
| 200 without the profile title (e.g. an interstitial) | Error: "Profile confirmation not found" |
| other status | Error: "Unexpected status: {code}" |

Existence is confirmed by the `<title>` rather than status alone, so a soft/interstitial 200 isn't misread as a hit. On a hit, the profile is enriched from the server-rendered markup with `name`, `user_id` (stable numeric account id), `country`, `gender`, `age`, `avatar`, and whatever activity stat counts the page renders (`videos`, `playlists`, `subscribers`, `achievements`, `likes`).

SpankBang sits behind **Cloudflare**, which fingerprints the **TLS handshake (JA3/JA4)** rather than the User-Agent. Python's default TLS stack (used by `httpx`) is now rejected with `403` regardless of the browser-like headers `make_request` sends — so a plain-`httpx` request no longer works.

The fix routes the request through `curl_cffi` with `impersonate="chrome"`, which replicates a real Chrome TLS handshake. Unlike a DataDome cookie wall, **no session warm-up is needed** here — Cloudflare's check on these GETs is passive fingerprinting, so the impersonated handshake alone passes:

| Approach | Result |
| --- | --- |
| `httpx` (default, or real Chrome UA + full headers) | 403 |
| `curl_cffi` `impersonate="chrome"` | **200** ✅ |
| `curl_cffi` `impersonate="safari18_0"` | 200 ✅ |
| `curl_cffi` `impersonate="chrome131"` / `"chrome136"` | 403 "Just a moment…" |

The impersonation target matters — the default `"chrome"` alias passes, while some pinned Chrome versions trip Cloudflare's interstitial — so the module uses the default.

This lives in a reusable core helper, `user_scanner/core/impersonate.py` (`impersonate_validate` / `impersonate_request`), that mirrors `generic_validate` but routes through a warmed, cookie-persistent impersonating session (the warm-up is optional — needed for cookie-gated walls like DataDome, skipped here). `curl_cffi>=0.7,<1` is added to `pyproject.toml`/`requirements.txt`, and to the lint job's install step so mypy type-checks against it.

## Notes

- **Username-only.** Email checking is not viable: registration (`/users/auth?ajax=1&register=1`) is reCAPTCHA-gated and exposes no async email-existence endpoint.
- Cloudflare is more volatile than a static header check and can escalate to active JS/managed challenges. It works cleanly today with the `"chrome"` target; if Cloudflare ever hard-challenges a given IP the module degrades to an error (403) rather than a wrong verdict.
- The `impersonate.py` helper is shared with the TripAdvisor module PR (#428) — the file is identical in both, so whichever merges first, the other's copy can be dropped from its diff.
- Neither holehe nor MailAccess has a SpankBang checker.

## Testing

- Verified live across 30+ existing handles → Found with metadata; random/non-existent → Not Found.
- `impersonate.py` has unit tests (`tests/test_impersonate.py`).

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama types-colorama
+          pip install httpx curl_cffi colorama types-colorama
           pip install ruff mypy
 
       - name: Run ruff
--- a/.github/workflows/lint-push.yml
+++ b/.github/workflows/lint-push.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama types-colorama
+          pip install httpx curl_cffi colorama types-colorama
           pip install ruff mypy
 
       - name: Run ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,8 @@ authors = [
 dependencies = [
   "httpx[http2]>=0.27,<0.29",
   "socksio>=1.0,<2",
-  "colorama>=0.4,<1"
+  "colorama>=0.4,<1",
+  "curl_cffi>=0.7,<1"
 ]
 
 requires-python = ">=3.10"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 httpx[http2]>=0.27,<0.29
 socksio>=1.0,<2
 colorama>=0.4,<1
+curl_cffi>=0.7,<1
--- a/tests/test_impersonate.py
+++ b/tests/test_impersonate.py
@@ -1,0 +1,89 @@
+from types import SimpleNamespace
+
+from user_scanner.core import impersonate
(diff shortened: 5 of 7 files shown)


brunolm:
~~Cloudflare / Age verification~~

kaifcodec:
@brunolm But it seems to be working atleast for only `Found` and `Not Found` checks, on my side!

kaifcodec:
Thanks @brunolm! Good call on introducing curl_cffi. Because these changes touch core files, I'll need a little extra time to review the PR carefully.
Till then feel free to add more high quality modules, these modules are really helping a lot, thanks again.

kaifcodec (approved):
@brunolm Thank you!
This is perfect now, merging it.

PR #423, opened 2026-07-19
feat: add Motherless email and username scan modules

brunolm:
## What

Adds **Motherless** to both scan modes — an email module and a username module.

> Note: the legacy `motherless.com` domain no longer resolves (NXDOMAIN). These modules target the site's current domain, **`motherless.xxx`**.

## How it works

Both use the registration form's field-availability endpoints (single unauthenticated POST each, no Turnstile required) which return a small HTML snippet:

**Email** — `POST /register/checkemail` (body: `email`):

| Response | Result |
|----------|--------|
| `class="available"` "email address is available" | Not Registered |
| `class="not-available"` "already used" | Registered |
| `class="not-available"` "Invalid Host." | Error (rejected) |

**Username** — `POST /register/checkusername` (body: `username`):

| Response | Result |
|----------|--------|
| `class="available"` "username is available" | Not Found |
| `class="not-available"` "Username not available." | Found |
| `class="not-available"` "Username is invalid." | Error (bad format) |

Both key off the `not-available`/`available` response class, with an "invalid" guard so rejected input isn't misreported as a hit. The username module's user-facing `show_url` points at the profile page `/m/{user}`.

## Notes

- These field-check endpoints are not behind the registration Turnstile, so a single request suffices — no cookie seeding needed.
- Signals are explicit ("already used" / "not available"), unlike the Aylo sites' ambiguous "requirements" heuristic.
- MailAccess's Sherlock list has a stale Motherless entry pointing at the dead `.com` domain; holehe has none.

## Testing

- Verified live against `motherless.xxx`: `brunolm7@gmail.com` → Registered and `brunolm` → Found; throwaway email/username → Not Registered / Not Found.

### DIFF
--- a/user_scanner/email_scan/adult/motherless.py
+++ b/user_scanner/email_scan/adult/motherless.py
@@ -1,0 +1,46 @@
+import httpx
+from user_scanner.core.result import Result
+
+
+async def _check(email: str) -> Result:
+    show_url = "https://motherless.xxx"
+    url = "https://motherless.xxx/register/checkemail"
+
+    headers = {
+        "User-Agent": "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Mobile Safari/537.36",
+        "X-Requested-With": "XMLHttpRequest",
+        "Origin": show_url,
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
LGTM.

PR #424, opened 2026-07-19
feat: add Chaturbate username scan module

brunolm:
## What

Adds a username-based **Chaturbate** check to `user_scan/adult/`.

## How it works

Chaturbate profile pages sit behind an age gate, so this uses the registration form's field-validation endpoint instead (an async module, since it needs the Django CSRF two-request flow):

1. `GET /accounts/register/` → extract `csrfmiddlewaretoken` (+ the `csrftoken` cookie).
2. `POST /accounts/ajax_validate_register_form/` with `username`.

| Response | Result |
|----------|--------|
| `errors.username` = "Username is already taken." | Found |
| no `username` error | Not Found |
| username format error (length / characters) | Error |

Field validation runs before the reCAPTCHA check, so no captcha token is needed. The user-facing `show_url` points at the profile page `/{user}/`.

## Notes

- **Username-only.** Email checking is not viable: the same endpoint returns only email *format/deliverability* errors ("Email is invalid", "Invalid email, please confirm") and never an already-registered signal, so email existence can't be enumerated. Password reset is behind the age gate.
- Sherlock and WhatsMyName both list Chaturbate as a username site (`/{user}/` profile), but that path now redirects to the age gate; the register-validate endpoint is a cleaner, explicit signal.
- holehe has no Chaturbate checker.

## Testing

- Verified live: `cute18cute`, `brunolm7` → Found; random valid names → Not Found; too-short names → Error.

### DIFF
--- a/user_scanner/user_scan/adult/chaturbate.py
+++ b/user_scanner/user_scan/adult/chaturbate.py
@@ -1,5 +1,6 @@
 import httpx
 import re
+from urllib.parse import unquote
 from user_scanner.core.result import Result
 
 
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
Perfect!

PR #425, opened 2026-07-19
feat: add Vimeo email scan module

brunolm:
Adds an email-existence scan module for Vimeo.

## How it works
- Fetches the `xsrft` token from `GET https://vimeo.com/_next/viewer`, then `POST https://vimeo.com/join` with `email_validation=true`.
- This is a validation-only call: no account is created, there is no captcha on this step, and no email is sent.

## Signal
- **Registered** → HTTP 418 `{"has_error_user_exists": true}` → `Result.taken`
- **Not registered** → HTTP 200 `[]` → `Result.available`

## Testing
- Auto-discovered (no registry changes). Full test suite passes (the only failures are the pre-existing `*_unreadable` tests that no-op on Windows).

### DIFF
--- a/user_scanner/email_scan/creator/vimeo.py
+++ b/user_scanner/email_scan/creator/vimeo.py
@@ -1,0 +1,59 @@
+import httpx
+from user_scanner.core.result import Result
+
+
+async def _check(email: str) -> Result:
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #426, opened 2026-07-19
feat: add Disqus email scan module

brunolm:
Adds an email-existence scan module for Disqus.

## How it works
- `GET https://disqus.com/profile/signup/` to grab the `csrfmiddlewaretoken`, then `POST` the same URL with the email, a **blank password**, and the required consent checkboxes.
- The blank password makes the form always fail validation, so **no account is ever created**. Django surfaces the email-uniqueness error before the reCAPTCHA check, so the signal is reachable unauthenticated (no captcha token needed).

## Signal
- **Registered** → body contains `The e-mail address you specified is already in use` → `Result.taken`
- **Not registered** → marker absent but the form still returns an `errorlist` (the blank-password error, confirming the form ran) → `Result.available`
- Neither → `Result.error` (guards against a block page / redesign silently reading as available)

## Testing
- Auto-discovered (no registry changes). Full test suite passes (only the pre-existing `*_unreadable` tests fail, and only on Windows).

### DIFF
--- a/user_scanner/email_scan/community/disqus.py
+++ b/user_scanner/email_scan/community/disqus.py
@@ -1,0 +1,59 @@
+import re
+import httpx
+from user_scanner.core.result import Result
+
+SIGNUP_URL = "https://disqus.com/profile/signup/"
+# Django surfaces this field error before the reCAPTCHA check, so an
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #427, opened 2026-07-19
feat: add Tumblr email scan module

brunolm:
Adds an email-existence scan module for Tumblr.

## How it works
- Scrapes the web app's `API_TOKEN` from the homepage, reads the `X-Csrf` header from `GET /api/v2/radar`, then `POST /api/v2/register/account/validate` with the email and a deliberately short password.
- This is a validation-only endpoint and the short password means **no account is ever created** and no email is sent.

## Signal
- **Registered** → `response.code == 2` ("User already exists") → `Result.taken`
- **Not registered** → `response.code == 1030` (password-too-short — the email cleared the uniqueness check) → `Result.available`
- Neither → `Result.error`

## Testing
- Auto-discovered (no registry changes). Full test suite passes (only the pre-existing `*_unreadable` tests fail, and only on Windows).

### DIFF
--- a/user_scanner/email_scan/social/tumblr.py
+++ b/user_scanner/email_scan/social/tumblr.py
@@ -1,0 +1,64 @@
+import re
+import httpx
+from user_scanner.core.result import Result
+
+# The public web app's bearer token, embedded in the homepage HTML.
+API_TOKEN_RE = re.compile(r'"API_TOKEN":"([^"]+)"')
+VALIDATE_URL = "https://www.tumblr.com/api/v2/register/account/validate"
+
+# response codes returned by the account-validate endpoint. A deliberately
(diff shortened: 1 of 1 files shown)


kaifcodec:
@brunolm I found an issue for which it was throwing,

`  [!] Tumblr (johndoe@gmail.com): Error (unexpected exception: <StreamReset stream_id:1, error_code:1, remote_reset:True>)`

- Under HTTP/2, this triggers a <StreamReset error_code:1> (REFUSED_STREAM).
- Under HTTP/1.1, the server abruptly closes the connection, leading to a Server disconnected without sending a response exception.

So I refactored it with `urllib` because its simpler TLS handshake signature and socket wrapping do not trigger the same bot-protection filters on Tumblr's servers.

kaifcodec (approved):

PR #428, opened 2026-07-19
feat: add TripAdvisor username scan module

brunolm:
Adds a username-existence scan module for TripAdvisor, plus a small reusable core helper for sites behind TLS-fingerprint bot walls.

## How it works
- `GET https://www.tripadvisor.com/Profile/{username}` to determine existence.
- **Found** → HTTP 200, or a 301/302 that redirects back to `/Profile/...` (TripAdvisor case-canonicalizes handles, e.g. `travelbug` → `/Profile/TRAVELBUG`, so a redirect to a profile URL still confirms the account exists).
- **Not found** → HTTP 404.
- On a hit, profile metadata (name, bio, hometown, join date, website, avatar) is enriched from TripAdvisor's persisted GraphQL query. Enrichment is best-effort and fully guarded, so a parsing/network failure never turns a found account into an error.

Since this module was first written, TripAdvisor moved behind **DataDome**, which fingerprints the **TLS handshake (JA3/JA4)** rather than the User-Agent. Python's default TLS stack (used by `httpx`) is rejected with `403` no matter which headers or cookies you send — so the original plain-`httpx` approach no longer works.

The fix needs two things **together**:
1. **`curl_cffi` with `impersonate="chrome"`** — replicates a real Chrome TLS handshake so the fingerprint check passes.
2. **A one-time session warm-up** — hitting the homepage first yields DataDome's clearance cookie, which the profile and GraphQL requests then reuse.

Neither alone is enough:

| Approach | Homepage | Profile | Works? |
| --- | --- | --- | --- |
| httpx + real Chrome UA + full headers, no warm-up | 403 | 403 | ❌ |
| httpx + session warm-up (cookies) | 403 | 403 | ❌ |
| curl_cffi chrome, no warm-up | — | 403 | ❌ |
| curl_cffi chrome + warm-up | 403 | 200 | ✅ |

This lives in a new reusable core helper — `user_scanner/core/impersonate.py` (`impersonate_validate` / `impersonate_request`) — that mirrors `generic_validate` but routes through a warmed, cookie-persistent impersonating session. Any other module that hits a DataDome/Cloudflare TLS wall can reuse it.

`curl_cffi>=0.7,<1` is added to `pyproject.toml` and `requirements.txt`.

## Testing
- Verified live against multiple real handles (`brunolm7`, `globetrotter`, `john`, `maria`, `david`, `susan`) → Found with correct metadata; well-formed nonexistent handles → Not Found.
- New `tests/test_impersonate.py` covers warm-up-once/session-reuse and error handling.
- Full suite passes (only the pre-existing `*_unreadable` tests fail, and only on Windows).

## Context
This fills a real coverage gap: Maigret and WhatsMyName check TripAdvisor usernames but user-scanner did not.

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama types-colorama
+          pip install httpx curl_cffi colorama types-colorama
           pip install ruff mypy
 
       - name: Run ruff
--- a/.github/workflows/lint-push.yml
+++ b/.github/workflows/lint-push.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama types-colorama
+          pip install httpx curl_cffi colorama types-colorama
           pip install ruff mypy
 
       - name: Run ruff
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ![User Scanner Logo](https://github.com/user-attachments/assets/49ec8d24-665b-4115-8525-01a8d0ca2ef4)
 <p align="center">
-  <img src="https://img.shields.io/badge/Version-1.4.1.7-blueviolet?style=for-the-badge&logo=github" />
+  <img src="https://img.shields.io/badge/Version-1.4.1.9-blueviolet?style=for-the-badge&logo=github" />
   <img src="https://img.shields.io/github/issues/kaifcodec/user-scanner?style=for-the-badge&logo=github" />
   <img src="https://img.shields.io/badge/Tested%20on-Termux-black?style=for-the-badge&logo=termux" />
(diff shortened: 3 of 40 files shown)


brunolm:
~~DataDome~~

kaifcodec:
@brunolm we had recently introduced -t and -C flag in PR #416 

One small thing I just noticed regarding timeout handling.

The new `impersonate.py` helper currently uses a hardcoded `DEFAULT_TIMEOUT = 15.0`, so it doesn't respect the global `-t` timeout flag that users can set from the CLI. This is a bit inconsistent with `make_request()` and `generic_validate()`, which already use `get_global_timeout()`.

Could you update both `impersonate_request()` and `_get_warm_session()` to do something like:

```python
global_timeout = get_global_timeout()
timeout = global_timeout if global_timeout is not None else DEFAULT_TIMEOUT
```

That way, modules such as SpankBang, TripAdvisor, and Pornhub will automatically honor the `-t` flag just like the rest of the codebase.

brunolm:
@kaifcodec that should be addressed now. I've also tested `-C 60` and it seems to work normally.

kaifcodec (approved):

PR #429, opened 2026-07-19
feat: add StackB username and email scan modules

kristoisberg:
Attempt number 3... 

Compared to the previous PR, I did some fuzz testing on the StackB module and only received unexpected responses for usernames longer than 16 characters, which is their limit. I added a check to interpret such usernames as available based on the received error message. Not sure about the timeouts though, the site seems to be available from everywhere.

I also added an email module, which works using the same login form as the username module.

I also found an interesting resource that could come handy for testing, containing all registered users: https://stackb.net/sitemap-profiles-1.xml (the numbers go up to 25).

### DIFF
--- a/user_scanner/email_scan/gaming/stackb.py
+++ b/user_scanner/email_scan/gaming/stackb.py
@@ -98,5 +98,4 @@ async def _check(email: str) -> Result:
 
 
 async def validate_stackb(email: str) -> Result:
-    """Check StackB's login response without sending email; repeated checks may be rate-limited."""
     return await _check(email)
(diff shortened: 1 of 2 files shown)


kaifcodec:
@kristoisberg did you test the email scan module? Do you know any email that is registered on StackB?

kaifcodec:
However I checked the username module, it is perfectly working now.

kristoisberg:
> @kristoisberg did you test the email scan module? Do you know any email that is registered on StackB?

I only know one registred email and I'd rather not disclose it, however I have tested it myself:
stackb

kaifcodec (approved):
@kristoisberg Thank you for the update and the PRs really added value to the project!
Will like to see you with more high quality PRs.

PR #430, opened 2026-07-19
feat: add osta.ee username module

kristoisberg:
Username module for the Estonian marketplace [osta.ee](https://osta.ee). It's a rather large one since it aggregates profile data from three pages.

Example company accounts can be acquired for testing from listings in this category: https://www.osta.ee/en/category/state-assets-and-enterprises and private accounts from other categories.


kaifcodec (approved):
@kristoisberg Again this is another high quality PR, it perfectly handles the data extraction and I like how you provide a link for the reference, excellent work on this on!

PR #433, opened 2026-07-22
feat: add Pornhub username scan module

brunolm:
## What

Adds a username-based **Pornhub** profile scan to `user_scan/adult/`, routed through a browser-impersonating request so it clears Pornhub's TLS-fingerprint bot block.

## How it works

Requests `https://www.pornhub.com/users/{user}`:

| Response | Result |
| --- | --- |
| 200 + `"<name>'s Profile - Pornhub.com"` title | Found (viewer) |
| 301/302 → `/users/<id>`, `/model/`, `/pornstar/`, `/channels/` | Found (renamed handle or promoted creator) |
| 404 | Not Found |
| other status | Error |

A direct 200 on `/users/<name>` is always a viewer account; models and pornstars 301-redirect to their own namespace, and a genuinely missing handle just 404s.

Pornhub returns **410 Gone** to Python's default TLS stack (`httpx`) regardless of headers — it fingerprints the **TLS handshake (JA3/JA4)**, not the User-Agent. Routing through `curl_cffi` with `impersonate="chrome"` replicates a real Chrome handshake and clears it. No session warm-up is needed (passive fingerprinting, unlike a DataDome cookie wall):

| Approach | Result |
| --- | --- |
| `httpx` (default, or real Chrome UA + full headers) | 410 |
| `curl_cffi` `impersonate="chrome"` | **200** ✅ |

This lives in a reusable core helper — `user_scanner/core/impersonate.py` (`impersonate_validate` / `impersonate_request`) — that mirrors `generic_validate` but routes through a browser-impersonating session. `curl_cffi>=0.7,<1` is added to `pyproject.toml`/`requirements.txt`, and to the lint job's install step so mypy type-checks against it.

## Metadata & account type

On a hit, the module follows the redirect (for models/pornstars) and scrapes the profile's About Me `infoPiece` block, then labels the account `type` from the resolved namespace:

- **viewer** → `name`, `user_id`
- **model** → `name`, `gender`, `profile_views`, `videos_watched`
- **pornstar** → full bio (`gender`, `relationship_status`, `birth_place`, `measurements`, `height`, `weight`, `ethnicity`, `star_sign`, `career_status`, `career_start_and_end`, …)

The `type` field follows the existing `dev/dockerhub.py` convention (`extra["type"]` = User/Organization) and makes the widely-varying `extra` self-describing — a consumer can tell a viewer from a verified pornstar directly instead of inferring it from which fields are present. The `infoPiece` extractor is generic, so whatever bio fields a profile fills in are captured automatically.

## Notes

- The `impersonate.py` helper is **shared with the TripAdvisor (#428) and SpankBang (#422) PRs** — the file is identical in all three, so whichever merges first, the later PRs' copies can be dropped from their diff.
- Redirect enrichment adds one extra request per model/pornstar hit; it is best-effort and never breaks detection.

## Testing

- Verified live: viewers (`brunolm`), models (`alex`), pornstars (`rileyreid` → `/pornstar/riley-reid`, 18-field bio), renamed handles (`mike` → `/users/9db2661`), and random → Not Found.

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama types-colorama
+          pip install httpx curl_cffi colorama types-colorama
           pip install ruff mypy
 
       - name: Run ruff
--- a/.github/workflows/lint-push.yml
+++ b/.github/workflows/lint-push.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install linters
         run: |
           python -m pip install --upgrade pip
-          pip install httpx colorama types-colorama
+          pip install httpx curl_cffi colorama types-colorama
           pip install ruff mypy
 
       - name: Run ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,8 @@ authors = [
 dependencies = [
   "httpx[http2]>=0.27,<0.29",
   "socksio>=1.0,<2",
-  "colorama>=0.4,<1"
+  "colorama>=0.4,<1",
+  "curl_cffi>=0.7,<1"
 ]
 
 requires-python = ">=3.10"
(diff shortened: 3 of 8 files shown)


kaifcodec (approved):
@brunolm Perfect!
Merging it!

PR #435, opened 2026-07-23
Handle ModelNotFoundException in Cropty validation

manav-py:
## Summary

This PR updates the Cropty validation logic to inspect the response body for `ModelNotFoundException` before returning an unexpected status error.

### What changed

- Preserved the existing behavior for `200` and `404` responses.
- Added a check for `ModelNotFoundException` in the response body for all other HTTP status codes.
- Return `Result.available()` when the API indicates the user does not exist, even if the response status is not explicitly handled.

### Why

Previously, only `200` and `404` responses were handled explicitly. Any other status code (such as `403`, `429`, or `500`) would immediately return:

```python
Result.error(f"Unexpected status: {response.status_code}")
```

without inspecting the response body.

With this change, if the API returns a `ModelNotFoundException` in the response body for an unexpected status code, it is correctly treated as an available username instead of an unexpected error.

### DIFF
--- a/user_scanner/user_scan/finance/cropty.py
+++ b/user_scanner/user_scan/finance/cropty.py
@@ -31,6 +31,20 @@ def validate_cropty(user):
         if response.status_code == 404:
             return Result.available()
 
+        
+        try:
+            data_json = response.json()
+
+            if (
+                "errors" in data_json
+                and data_json["errors"]
+                and data_json["errors"][0].get("code") == "ModelNotFoundException"
+            ):
+                return Result.available()
+
+        except Exception:
+            pass
+        
         return Result.error(f"Unexpected status: {response.status_code}")
 
     return generic_validate(url, process, headers=headers, show_url=show_url)


kaifcodec:
@mnv-cs You were spot on here as well! However, the approach in #436 turned out to be a bit cleaner. Another blocker was that the PR author and the commit author didn't match, which goes against our repository's contribution guidelines. 

We really appreciate your time and effort on this though! You're always welcome to contribute again, whether that's adding new modules or fixing bugs. We'd be happy to review and merge your future PRs, just make sure your GitHub account matches the commit author next time.

manav-py:
Thanks for the review and for the detailed feedback! I appreciate you
taking the time to explain both the implementation preference and the
commit author issue. I've fixed my Git configuration so future commits will
use the correct author information. Looking forward to contributing again!

kaifcodec:
No worries @mnv-cs !
You can join our User Scanner discord server from README badges, if you want to keep track on new issues and contribute further to `user-scanner`.

PR #437, opened 2026-07-23
fix: reject usernames causing false positives in Calendly, IFTTT, OpenStreetMap modules

kristoisberg:
This PR rejects usernames containing periods in Calendly, IFTTT, and OpenStreetMap, which were causing false positives as websites were only checking the part of username before the period. A good test case is the username "test.ddfdddgfdgf".

### DIFF
--- a/user_scanner/user_scan/other/calendly.py
+++ b/user_scanner/user_scan/other/calendly.py
@@ -3,7 +3,7 @@ from user_scanner.core.orchestrator import Result, make_request
 
 def validate_calendly(user):
     if "." in user:
-        return Result.error("Username cannot contain periods")
+        return Result.available("Username cannot contain periods")
 
     api_url = f"https://calendly.com/api/booking/profiles/{user}"
     show_url = f"https://calendly.com/{user}"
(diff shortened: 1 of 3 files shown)


kaifcodec:
@kristoisberg I will review it after some hours, thank you for the PR.

user_scanner/user_scan/other/calendly.py
```diff
@@ -2,6 +2,9 @@
 from user_scanner.core.orchestrator import Result, make_request
 
 def validate_calendly(user):
+    if "." in user:
+        return Result.error("Username cannot contain periods")
```
kaifcodec: @kristoisberg I see where you are coming from, but i think instead of using `Result.error()` we should use `Result.available()` and pass that exact same message so logically we are still correct that is the username doesn't exist there on top of that the message provides the reason.
```suggestion
        return Result.available("Username cannot contain periods")
```

kaifcodec (changes requested):

kaifcodec (dismissed):
@kristoisberg Thank you for catching this! 

Great work on this. If you stay consistent and have dedicated time for the project, we would definitely be open to bringing you on as a maintainer down the line, specifically to help manage error handling, site updates, and resolving false positives or negatives alongside the team. Keep it up!

kaifcodec (approved):

PR #438, opened 2026-07-23
fix: handle Freesound rate limited responses as errors instead of positives

kristoisberg:
Freesound responds with 403 when you are rate limited. The module currently handles these responses as positives...

### DIFF
--- a/user_scanner/user_scan/music/freesound.py
+++ b/user_scanner/user_scan/music/freesound.py
@@ -12,6 +12,9 @@ def validate_freesound(user):
     }
 
     def process(response):
+        if response.status_code == 403:
+            return Result.error("Freesound rate limit or blocked")
+
         if response.status_code == 404 or "Page not found" in response.text:
             return Result.available()
 
--- a/user_scanner/user_scan/music/freesound.py
+++ b/user_scanner/user_scan/music/freesound.py
@@ -13,7 +13,7 @@ def validate_freesound(user):
 
     def process(response):
         if response.status_code == 403:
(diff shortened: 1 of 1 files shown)


user_scanner/user_scan/music/freesound.py
```diff
@@ -12,6 +12,9 @@ def validate_freesound(user):
     }
 
     def process(response):
+        if response.status_code == 403:
+            return Result.error("Freesound rate limit or blocked")
```
kaifcodec: @kristoisberg This error message is much cleaner!

However, if the Freesound module is consistently returning 403 status codes across different regions and accounts, the endpoint may have updated its request validation or security checks. If you are able to inspect the network requests from a real browser session and update the headers/request parameters to match what the site currently expects, that would make a great follow-up PR! Let us know what you observe when testing the live endpoint.

```suggestion
            return Result.error("Freesound rate limit or blocked with status code 403")

kaifcodec (changes requested):

user_scanner/user_scan/music/freesound.py
kristoisberg: This was a case of rate limiting as it started happening after iterating over a considerable number of username permutations. I am able to access the site in normal circumstances, but my IP was blocked for some time as both user-scanner and my browser couldn't access it for a while.

kaifcodec: @kristoisberg Okay in that case just applying the changes and merging it.
Thank you for the information.

PR #439, opened 2026-07-23
feat: add XVideos username scan module

brunolm:
## What

Adds a username-based **XVideos** profile scan to `user_scan/adult/`.

## How it works

Requests `https://www.xvideos.com/profiles/{user}`:

| Response | Result |
| --- | --- |
| 200 + `"<name> - Profile page - XVIDEOS.COM"` title | Found |
| 404 ("Unknown profile") | Not Found |
| other status | Error |

Unlike the recent adult modules, **no `curl_cffi` is needed** — XVideos serves `/profiles/` over plain nginx with no TLS-fingerprint bot wall, so this uses the standard httpx `generic_validate` path. `/profiles/` is also the canonical namespace: models/channels **redirect into** it (`/models/mia → /profiles/mia`), so a single lookup covers regular users and creators alike.

## Metadata

Identity comes from the page's embedded JSON `user` object (`name`, `user_id`, `avatar`), parsed with `json.loads` so escapes resolve cleanly. Everything else is scraped from the visible profile-info panel with a generic row extractor:

```
<p id="pinfo-<field>"><strong>Label:</strong><span>Value</span></p>
```

This captures whatever a profile fills in — no per-field regex to maintain — so it scales with the account: gender, age, country, languages, signup date, last activity, plus optional personal/physical details (relationship, education, religion, ethnicity, height, weight, …). Verified live: a rich viewer profile yielded 20+ fields, a model profile ~22 (with `profile_hits`, `subscribers`), and a nonexistent handle → Not Found.

## Notes

- No account-`type` field: unlike Pornhub (where the redirect namespace reveals viewer/model/pornstar), XVideos serves everything under `/profiles/` with no reliable in-page classifier, so I left `type` out rather than guess.

## Testing

- Verified live: existing handles → Found with metadata; random/non-existent → Not Found.

### DIFF
--- a/user_scanner/user_scan/adult/xvideos.py
+++ b/user_scanner/user_scan/adult/xvideos.py
@@ -10,14 +10,18 @@ def validate_xvideos(user):
     show_url = url
 
     def process(response):
+        # A genuinely missing handle 404s with an "Unknown profile" page (that
+        # phrase is its <title>); require the marker so a transient or blocked
+        # 404 isn't misread as "available".
         if response.status_code == 404:
-            return Result.available(url=show_url)
+            if "unknown profile" in response.text.lower():
+                return Result.available(url=show_url)
+            return Result.error("Unexpected 404 (not the profile-not-found page)", url=show_url)
 
         if response.status_code != 200:
             return Result.error(f"Unexpected status: {response.status_code}", url=show_url)
 
-        # A real profile is titled "<name> - Profile page - XVIDEOS.COM"; a
-        # missing handle 404s with an "Unknown profile" page.
+        # A real profile is titled "<name> - Profile page - XVIDEOS.COM".
         if "profile page" in _title(response.text).lower():
             return Result.taken(extra=_extract_profile(response.text), url=show_url)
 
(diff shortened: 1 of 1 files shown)


user_scanner/user_scan/adult/xvideos.py
```diff
@@ -0,0 +1,72 @@
...
+
+from user_scanner.core.orchestrator import generic_validate, Result
+
+
+def validate_xvideos(user):
+    url = f"https://www.xvideos.com/profiles/{user}"
+    show_url = url
+
+    def process(response):
+        if response.status_code == 404:
```
kaifcodec, line 16: @brunolm Can you find any other explicit string that the website might return in the response payload, like, "user not found", so we can use that in 
```python
if response.status_code == 404 and "user not found" in response.text:
```
 block that would keep it safe from false negative.

if it doesn't return something like that we can stick to it.

brunolm: Updated

kaifcodec:
@brunolm just heads-up that #440 has been merged you can sync this branch now to get the checks passed!

kaifcodec (approved):
@brunolm This is perfect now and I tested with usernames that I know the results of, and it worked perfectly. Merging it.

PR #441, opened 2026-07-24
feat: add XNXX username scan module

brunolm:
Adds a username scan module for XNXX.

**Namespaces**
- `/pornstar/<user>` serves both models and uploaders; the account `type` is derived from the embedded `user.url` (`/pornstar/` → model vs `/porn-maker/` → porn_maker), not from which path answered.
- Plain members have no public profile anywhere, so existence is confirmed via the signup form's live name check: `GET /account/checkprofilename?profile=<name>` (read-only, sends no email). This is the authoritative signal; the profile page is fetched only to enrich.

**Correctness**
- Not-found is confirmed by the explicit "Unknown profile" title, never a bare 404.
- The check endpoint's `code:1` is overloaded (taken vs "too short"/"not allowed"), so the verdict is taken from the message, and invalid handles return `error`, not a verdict.

**Metadata**: name, numeric id, gender, avatar, aliases (+ account type).

Transport: curl_cffi (browser impersonation). Tested live against real and nonexistent handles on each namespace. `ruff check .` and `mypy user_scanner` pass clean.

### DIFF
--- a/user_scanner/user_scan/adult/xnxx.py
+++ b/user_scanner/user_scan/adult/xnxx.py
@@ -1,0 +1,118 @@
+import re
+import json
+import html
+
+from user_scanner.core.impersonate import impersonate_request, impersonate_validate
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.xnxx.com"
+
+# The signup form's live name check. It answers for every registered account,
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #442, opened 2026-07-24
feat: add YouPorn username scan module

brunolm:
Adds a username scan module for YouPorn.

**Namespaces** (a handle lives in exactly one, so all are checked before reporting available):
- `/pornstar/<user>/` — listed models
- `/amateur/<user>/` — verified uploaders
- `/uservids/<user>/` — plain accounts; canonicalises to `/user/<numeric id>/`

`/user/<name>` is deliberately **not** used — it is a soft-200 catch-all that renders the generic home page for any string (the handle appears only in canonical/hreflang boilerplate).

**Correctness**
- Found is confirmed by the canonical link echoing the requested handle (the host varies because YouPorn geo-redirects, e.g. `br.youporn.com`).
- Not-found is confirmed by the "404 Page Not Found" title.

**Metadata**: name, rank/views/subscribers, avatar, and the bio table. Note bio-table keys arrive translated (YouPorn geo-localises and honours no `lang` override); values are unaffected.

Transport: curl_cffi, following the geo 302. Tested live against real (model/uploader/viewer) and nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/youporn.py
+++ b/user_scanner/user_scan/adult/youporn.py
@@ -1,0 +1,131 @@
+import re
+import html
+
+from user_scanner.core.impersonate import impersonate_request
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.youporn.com"
+
+# Accounts are split across namespaces and a handle sits in only one, so every
+# namespace must confirm the miss before a name counts as free. Note /user/ is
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@brunolm Perfectly done.
LGTM.

PR #443, opened 2026-07-24
feat: add Stripchat username scan module

brunolm:
Adds a username scan module for Stripchat.

**Namespaces**
- `/<user>` — model pages (200, embedded `model` object in `__PRELOADED_STATE__`).
- Viewer accounts: `/<user>` 302-redirects to `/user/<user>` only when the account exists (unknown handles 404). The redirect is the whole signal — that page is client-rendered — so `allow_redirects=False` and the redirect target is the verdict.

**Correctness**
- Not-found confirmed by the embedded `"type":"notFound"` error object (the 404 page still echoes the handle).
- The embedded model object is matched on `username` rather than taken first, since recommendation carousels embed sibling `"model":{…}` nodes.

**Metadata**: models get the full profile from the embedded state; viewers are enriched from the front API the page itself calls (all anonymous): `/api/front/users/user-ids/<h>` → id, then `/api/front/v2/users/<id>[/profile]` → username, country, bio, interests, etc. Unset numeric fields (age 0) are dropped, keeping real booleans.

Transport: curl_cffi. Tested live against models and viewer accounts plus nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/stripchat.py
+++ b/user_scanner/user_scan/adult/stripchat.py
@@ -1,0 +1,187 @@
+import re
+import json
+
+from user_scanner.core.impersonate import impersonate_request, impersonate_validate
+from user_scanner.core.result import Result
+
+BASE_URL = "https://stripchat.com"
+API_URL = f"{BASE_URL}/api/front"
+
+# Profile fields worth surfacing, mapped to friendlier extra keys.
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #444, opened 2026-07-24
feat: add LiveJasmin username scan module

brunolm:
Adds a username scan module for LiveJasmin.

**Namespaces**
- `/en/chat/<user>` — performers. Both states answer 200, so the body decides: a real performer page carries a `performerName` meta tag (which persists while the model is offline, so it is an existence signal, not a live-stream one). The miss page is titled "<handle> is not available".
- Members have no public page at all; existence is confirmed via the signup form's live nick check: `POST /en/auth/validate/check-nick` with `nick` + a `csrfToken` scraped from the sign-up page (read-only, sends no email).

**Correctness / ordering**
- Performers must be checked first: the nick check reports performers as *available*, since they live in a separate namespace.
- The nick check's `errors` field is overloaded (also "too short"/"too long"), so the verdict is taken from the message; usernames are 6–16 chars, so out-of-range handles return `error`, not a verdict.

**Metadata**: performer name, id, category, avatar (persistent). Members return `type: member`.

Transport: curl_cffi. Tested live against online + offline performers and member accounts plus nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/livejasmin.py
+++ b/user_scanner/user_scan/adult/livejasmin.py
@@ -1,0 +1,142 @@
+import re
+import json
+import html
+import threading
+
+from user_scanner.core.impersonate import impersonate_request
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.livejasmin.com"
+SIGNUP_URL = f"{BASE_URL}/en/auth/sign-up"
+# The signup form's live nick check. Performers are a separate namespace and read
+# as free here, so it only settles plain member accounts.
+CHECK_URL = f"{BASE_URL}/en/auth/validate/check-nick"
+CHECK_HEADERS = {
+    "X-Requested-With": "XMLHttpRequest",
+    "Referer": SIGNUP_URL,
+    "Origin": BASE_URL,
+}
+
+_token_lock = threading.Lock()
+_form_token: str | None = None
+
+# Fields from the embedded player config, mapped to friendlier extra keys.
+CONFIG_FIELDS = {
+    "display_name": "name",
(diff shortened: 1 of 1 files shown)


kaifcodec:
It looks like the module is hitting a DNS resolution failure (`curl: (6) could not resolve host: www.livejasmin.com`). 

This usually means either the base domain/subdomain URL configured in the module is incorrect or the platform uses a different active domain/endpoint now. Could you double-check the URL structure in the module to ensure `www.livejasmin.com` is the correct host being queried?

Raw error:
`  [!] Livejasmin [https://www.livejasmin.com/en/chat/johndoe] (johndoe): Error (DNSError: Failed to perform, curl: (6) could not resolve host: www.livejasmin.com. see https://curl.se/libcurl/c/libcurl-errors.html first for more details.)
`

brunolm:
Confirmed https://www.livejasmin.com is the correct host.

Off-VPN the module works (e.g. brunolm7 → Found, type member).

On a VPN that fails DNS for this domain I get the same error you reported:
`curl: (6) could not resolve host: www.livejasmin.com.`

So, this is environment/DNS, not a wrong URL. DNS failures correctly return `Result.error` and don’t emit a free/taken verdict.

# no VPN
image


# VPN
image

---

Happy to close this PR if needed

kaifcodec (approved):
@brunolm Yeah, you're exactly right. I just verified with nslookup that the domain resolves perfectly fine when using Cloudflare (1.1.1.1). The issue is definitely local DNS poisoning/hijacking from my provider/VPN.

MERGING IT!

PR #445, opened 2026-07-24
feat: add CamSoda username scan module

brunolm:
Adds a username scan module for CamSoda.

**Namespaces**
- `/<user>` — broadcasters (200, embedded `user` object with rich profile).
- Members have no public page; existence is confirmed via the signup availability check: `GET /api/v1/user/check?type=username&value=<h>` → `{"match":1}` taken / `{"match":0}` free (read-only, anonymous, covers both account types).

**Correctness**
- The check is case-insensitive and whitespace-trimmed, matching the site.
- `match:1` is overloaded for sub-3-char reserved names (reports taken with no account behind it), so handles under 3 characters return `error`, not a verdict.
- The broadcaster `user` object is matched on `username` rather than taken first (recommendation carousels embed siblings).

**Metadata**: broadcasters get name, id, gender, created date, followers, rating, tags, avatar; members return `type: member`.

Transport: curl_cffi. Tested live against broadcasters and member accounts plus nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/camsoda.py
+++ b/user_scanner/user_scan/adult/camsoda.py
@@ -1,0 +1,147 @@
+import re
+import json
+
+from user_scanner.core.impersonate import impersonate_request
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.camsoda.com"
+# The signup form's live availability check. It answers for every account,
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #446, opened 2026-07-24
feat: add Cam4 username scan module

brunolm:
Adds a username scan module for Cam4.

**Approach**
- `GET /rest/v1.0/profile/<user>/info` — a clean JSON API that covers every account type (broadcasters and plain members alike), so no separate member handling is needed.
- Found is confirmed by the payload's `username` matching the request; a missing handle returns 404 with an empty body.

**Correctness**
- Unset numeric fields come back as `0` (e.g. `age` when no birth date is set) and are dropped, while real booleans (`online: false`) are preserved.
- Enum values that repeat their own field name are normalised.

**Metadata**: id, age, gender, ethnicity, marital status, location, physical attributes, languages, body decorations, created/last-broadcast dates, social links, wishlists, bio.

Transport: curl_cffi. Tested live against multiple accounts plus nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/cam4.py
+++ b/user_scanner/user_scan/adult/cam4.py
@@ -1,0 +1,118 @@
+import json
+import re
+from datetime import datetime, timezone
+
+from user_scanner.core.impersonate import impersonate_validate
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.cam4.com"
+
+# Scalar profile fields worth surfacing, mapped to friendlier extra keys.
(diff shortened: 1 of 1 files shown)


kaifcodec:
@brunolm I just tested this module across a wide range of usernames, but it is currently throwing unexpected status code errors for every check.

For example:
- `blue` and `johndoe` -> 200 OK
- `jason` -> 403
- `alex` -> 401

Could you re-check this on your end?

PR #447, opened 2026-07-24
feat: add Erome username scan module

brunolm:
Adds a username scan module for Erome.

**Approach**
- `/<user>` — a single public profile namespace. Found is confirmed by the "<handle> - Profile page - EroMe" title matching the request; not-found by the "Page not found" heading.

**Correctness**
- A removed/DMCA'd account answers **410** with a "User deleted" page — that name is neither free nor backed by a live profile, so it returns `error`, never a verdict (mirrors the Erome/410 case noted in CONTRIBUTING).

**Metadata**: display name, numeric account id (from the report control — the only stable id, since the handle can change), album/view/follower/following counts (zero counts dropped), and account seniority.

Transport: curl_cffi. Tested live against real profiles, a deleted account, and nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/erome.py
+++ b/user_scanner/user_scan/adult/erome.py
@@ -1,0 +1,80 @@
+import re
+import html
+
+from user_scanner.core.impersonate import impersonate_validate
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.erome.com"
+
+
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #448, opened 2026-07-24
feat: add e621 username scan module

brunolm:
Adds a username scan module for e621.

**Approach**
- `GET /users/<user>.json` — e621's public REST API. Found is confirmed by the JSON `name` matching the request (the endpoint also resolves numeric ids, which are rejected here); a missing account returns 404 with `{"success":false,"reason":"not found"}`.
- Sends a descriptive `User-Agent` as e621's API terms require; generic/spoofed browser agents get throttled.

**Metadata**: id, level, join date, verified/banned flags, activity counters (uploads, edits, comments, favourites, feedback — non-zero only), avatar post link, and profile about/artist-info text.

Transport: curl_cffi with the required descriptive UA. Tested live against real accounts and nonexistent handles. `ruff` + `mypy` clean.

### DIFF
--- a/user_scanner/user_scan/adult/e621.py
+++ b/user_scanner/user_scan/adult/e621.py
@@ -1,0 +1,103 @@
+import json
+import re
+
+from user_scanner.core.impersonate import impersonate_validate
+from user_scanner.core.result import Result
+
+BASE_URL = "https://e621.net"
+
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
LGTM.

PR #450, opened 2026-07-24
Fix: replace silent Result.taken() fallthrough with explicit Result.error()

### LINKED ISSUE #449: [Bug / Refactor] Silent Result.taken() fallthrough on malformed 200 responses across multiple modules
## Description

Several `user_scan` validators silently swallow parse failures on a `200` response and fall through to an unconditional `Result.taken()`, instead of returning `Result.error(...)` as required by the module guidelines:

> **Strict Error Handling:** NEVER use `raise Exception()`. All unhandled states or unexpected status codes must return `Result.error(f"Unexpected status code {resp.status_code}")`.
> — CONTRIBUTING.md

When `response.json()` fails, or the parsed body doesn't have the expected keys, these modules report the username as **taken** rather than surfacing the failure — which can produce false positives whenever the API returns a malformed, rate-limited, or otherwise unexpected `200` body.

This is the same bug class fixed in `cropty.py` via #436 — this issue tracks the same pattern found elsewhere in the codebase.

## Affected Files

| File | Pattern |
|---|---|
| `user_scanner/user_scan/finance/destream.py` | `except Exception: pass` → falls through to `return Result.taken()` |
| `user_scanner/user_scan/gaming/apexlegends.py` | same |
| `user_scanner/user_scan/gaming/minecraft.py` | same |
| `user_scanner/user_scan/gaming/lichess.py` | same |
| `user_scanner/user_scan/gaming/chess_com.py` | same |
| `user_scanner/user_scan/dev/codecademy.py` | `except Exception: return Result.taken()` directly (any exception at all is reported as "taken") |
| `user_scanner/user_scan/dev/githubgist.py` | same as codecademy |

## Example (`destream.py`)

```python
def process(response):
    if response.status_code == 200:
        try:
            data = response.json()
            ...
            return Result.taken(extra=extra)
        except Exception:
            pass
        return Result.taken()   # ← silent fallthrough, should be Result.error(...)

    if response.status_code == 404:
        return Result.available()

    return Result.error(f"Unexpected status: {response.status_code}")
```

## Proposed Fix

Replace the bare fallback `return Result.taken()` (and the `except Exception: return Result.taken()` variant) in each file with an explicit `Result.error(...)`, e.g.:

```python
except Exception:
    pass
return Result.error("Unexpected 200 response shape (no recognizable data)")
```

---

I'm happy to open a PR for this myself. Could I get assigned?

A-S-Manoj:
Closes #449

## Summary

Several `user_scan` validators silently swallowed parse failures or unrecognized `200` response shapes and fell through to an unconditional `Result.taken()`, instead of returning `Result.error(...)` as required by the module guidelines (Strict Error Handling section of CONTRIBUTING.md).

This replaces that fallback with an explicit error across all 7 affected files:

```python
return Result.error("200 response status with no recognizable data, report it via GitHub issues")
```

## Files changed

- `user_scanner/user_scan/finance/destream.py`
- `user_scanner/user_scan/gaming/apexlegends.py`
- `user_scanner/user_scan/gaming/minecraft.py`
- `user_scanner/user_scan/gaming/lichess.py`
- `user_scanner/user_scan/gaming/chess_com.py`
- `user_scanner/user_scan/dev/codecademy.py`
- `user_scanner/user_scan/dev/githubgist.py`

## Testing

- `ruff check .` — passed
- `mypy user_scanner` — passed
- `pytest` — no test regressions (no per-module tests exist for these validators; existing suite unaffected)

### DIFF
--- a/user_scanner/user_scan/dev/codecademy.py
+++ b/user_scanner/user_scan/dev/codecademy.py
@@ -37,7 +37,7 @@ def validate_codecademy(user):
                     return Result.taken(extra={"is_private": False})
                 return Result.available()
             except Exception:
-                return Result.taken()
+                return Result.error("200 response status with no recognizable data, report it via GitHub issues")
 
         return Result.error("Unexpected response body, report it via GitHub issues.")
 
--- a/user_scanner/user_scan/dev/githubgist.py
+++ b/user_scanner/user_scan/dev/githubgist.py
@@ -25,7 +25,7 @@ def validate_githubgist(user):
                 
                 return Result.taken(extra=extra)
             except Exception:
-                return Result.taken()
+                return Result.error("200 response status with no recognizable data, report it via GitHub issues")
                 
         elif response.status_code == 403:
             return Result.error("Rate limited by GitHub API")
--- a/user_scanner/user_scan/finance/destream.py
+++ b/user_scanner/user_scan/finance/destream.py
@@ -26,7 +26,7 @@ def validate_destream(user):
                 return Result.taken(extra=extra)
             except Exception:
                 pass
-            return Result.taken()
+            return Result.error("200 response status with no recognizable data, report it via GitHub issues")
 
         if response.status_code == 404:
             return Result.available()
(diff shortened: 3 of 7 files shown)


kaifcodec (approved):
@A-S-Manoj Thanks for the PR, it is ok to me. Merging it.

PR #451, opened 2026-07-24
Add CodePen username availability module (dev category)

SpiliosDimakopoulos:
Adds a new username availability check for CodePen (codepen.io), following the existing dev/ module pattern.

- File: user_scanner/user_scan/dev/codepen.py
- Exposes validate_codepen(user) using generic_validate
- Available: HTTP 404 on profile URL
- Taken: HTTP 200 + explicit match on og:url meta tag or "(@username)" in page title (avoids false positives from bare 200 status)
- Error: any other unexpected status code

Tested locally with a known existing username and a random non-existent username to confirm both Available/Taken paths resolve correctly.

No changes needed elsewhere — module is picked up automatically by the existing discovery mechanism.

### DIFF
--- a/user_scanner/user_scan/dev/codepen.py
+++ b/user_scanner/user_scan/dev/codepen.py
@@ -1,0 +1,21 @@
+from user_scanner.core.orchestrator import generic_validate, Result
+
+def validate_codepen(user):
+    url = f"https://codepen.io/{user}"
+    show_url = url
+
+    def process(response):
+        if response.status_code == 404:
+            return Result.available()
+
+        if response.status_code == 200:
+            marker_og = f'og:url" content="https://codepen.io/{user}'
+            marker_title = f'(@{user})'
+            if marker_og in response.text or marker_title in response.text:
+                return Result.taken(extra={"profile": url})
+            return Result.error("Ambiguous response body, report it via GitHub issues.")
+
+        return Result.error(f"Unexpected status code {response.status_code}")
+
+    headers = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"}
+    return generic_validate(url, process, show_url=show_url, headers=headers)


user_scanner/user_scan/dev/codepen.py
```diff
@@ -0,0 +1,21 @@
+from user_scanner.core.orchestrator import generic_validate, Result
+
+def validate_codepen(user):
```
A-S-Manoj, line 3: Flagging missing type hints per CONTRIBUTING.md convention:

```suggestion
def validate_codepen(user: str) -> Result:
```

A-S-Manoj (commented):
Tested across 4 rounds on two networks (home wifi, mobile hotspot). Cloudflare (`cf-mitigated: challenge`, title `"Just a moment..."`) blocked the request in the large majority of attempts, including 4/4 attempts against the "available" (nonexistent-user) path, and 3/4 against the "taken" (existing-user) path. One request did get through with a real 200, but the very next identical request (same network, same headers) was challenged again seconds later. This means results are non-deterministic in practice: the same username can resolve correctly one run and silently fall through to `Result.error(...)` on the next, with no retry logic in the module to smooth this out. Given the failure rate observed, a user of the tool would likely see `error` for CodePen far more often than a real `available`/`taken` result. Worth considering a Cloudflare-resistant request approach (browser-impersonation headers/TLS fingerprinting, or an API endpoint) before this is reliable.

kaifcodec:
@A-S-Manoj I am not sure about the PR author's activity but most likely they aren't going to work on it,  It looks like AI generated PR (open and leave type).

If you want you can work on it and open a new PR on this using our existing `impersonate_validate()` function (check PR #422 for reference) if that works for this module.

Anyways what about #463 ? Are you working on that?

A-S-Manoj:
@kaifcodec Thanks for the context. I tested this yesterday and retested today morning across two different networks, so I can confirm the problem isn't about rate limiting from my testing, that's why I reviewed this first before responding on the issue thread.

My schedule is a bit tight right now. Let's give the author a bit more time to respond. If there's no response, maybe we can post it as an issue and someone interested can work on it with your guidance.

On #463, yes I'm actively working on it. Plan is locked in from the discussion there (TTY-aware y/n/d prompt with config persistence), I'll open a PR once it's ready.

kaifcodec:
**WILL BE COVERED IN #452 IF MERGED. SO CLOSING IT AS DUPLICATE.**

PR #452, opened 2026-07-24
Add Lobsters (lobste.rs) and Codepen username availability module

SpiliosDimakopoulos:
Adds a username availability check for Lobsters (lobste.rs), a tech/OSINT community site currently missing from user_scan/community/.

- File: user_scanner/user_scan/community/lobsters.py
- Uses the public, documented JSON endpoint lobste.rs/u/<username>.json
- Available: HTTP 404
- Taken: HTTP 200 + parsed JSON, extracts about/karma/created_at/github_username when present
- Error: any other status code or unparseable body

Tested locally against a known existing username and a random non-existent one; both paths resolve correctly.

### DIFF
--- a/user_scanner/user_scan/dev/codepen.py
+++ b/user_scanner/user_scan/dev/codepen.py
@@ -1,0 +1,21 @@
+from user_scanner.core.orchestrator import generic_validate, Result
+
+def validate_codepen(user):
+    url = f"https://codepen.io/{user}"
+    show_url = url
+
+    def process(response):
+        if response.status_code == 404:
+            return Result.available()
+
+        if response.status_code == 200:
+            marker_og = f'og:url" content="https://codepen.io/{user}'
+            marker_title = f'(@{user})'
+            if marker_og in response.text or marker_title in response.text:
+                return Result.taken(extra={"profile": url})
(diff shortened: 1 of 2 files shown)


user_scanner/user_scan/dev/codepen.py
```diff
@@ -0,0 +1,21 @@
+from user_scanner.core.orchestrator import generic_validate, Result
+
+def validate_codepen(user):
+    url = f"https://codepen.io/{user}"
+    show_url = url
+
+    def process(response):
+        if response.status_code == 404:
```
kaifcodec, line 8: @SpiliosDimakopoulos Could you check if the response payload contains any other explicit strings like `"user doesn't exist"` or `"profile not found"`? We could use a targeted check like:

```python
if response.status_code == 404 and "profile not found" in response.text:
```

this removes the risk of any false negatives in future.

kaifcodec:
@SpiliosDimakopoulos Another issue I noticed is that the CodePen module is returning a 403 status code on my machine. This could be due to WAF restrictions that require custom headers, cookies, or updated request handling. Could you re-test this module on your end and verify if it requires updated headers?

PR #454, opened 2026-07-24
docs: fix incorrect example output for {n;m} pattern syntax

SpiliosDimakopoulos:
The {1;3} example in the Length Control section showed "useraa" (a
2-character expansion) in its output list. But {n;m} is documented (and
implemented in patterns.py's _parse_lenset) as "exactly n or m characters"
— not a range — so length-2 results are never produced for {1;3}.

Verified directly:
>>> list(expand_patterns("user[a-c]{1;3}"))
# 30 results total: 3 of length 1, 27 of length 3, no length-2 entries

Fixed the example to show a length-3 result ("useraaa") instead of the
impossible length-2 one. Docs only, no code changes.


kaifcodec (approved):
@SpiliosDimakopoulos Good catch here! Thank you for the fix.

PR #456, opened 2026-07-24
Add missing test coverage for status_validate and generic_validate

SpiliosDimakopoulos:
core/orchestrator.py's status_validate was only tested for the "available" path. This adds coverage for the remaining branches:

- taken status match
- no-match status (falls through to Result.error)
- ambiguous status (same code present in both available/taken lists -> Result.error, can't be both)
- list-based available/taken codes (not just single ints)
- generic_validate exception handling (make_request raising -> Result.error with correct url)

No behavior changes, test-only PR. All 9 tests in tests/test_orchestrator.py pass locally.


kaifcodec:
**MISMATCH WITH THE PR TITLE AND ACTUAL COMMIT.**

PR #465, opened 2026-07-26
fix: respect --allow-loud for -m/--module scans

### LINKED ISSUE #463: [Bug] `-m/--module` silently bypasses the `--allow-loud` safety filter
**Description:**
`--allow-loud` exists so users explicitly opt in to scanning sites that notify the target being scanned (password reset emails, etc, see `LOUD_MODULES` in `helpers.py`). But when scanning via `-m/--module`, that opt-in gets silently overridden:

```python
# user_scanner/__main__.py
if args.module:
    fn = run_email_module_batch if is_email else run_user_module
    module_config = replace(config, allow_loud=True)   # forced True, unconditionally
    for module in validated_modules:
        results.extend(fn(module, target, module_config))
```

Running `user-scanner -u someone -m netflix` scans a loud module and can trigger a real notification to the target, even without `--allow-loud`, with no warning printed. This is inconsistent with the `-c/--category` path a few lines below, which correctly passes the user's actual `config` and respects `allow_loud` as set.

For a recon/OSINT tool, this undermines the one guarantee `--allow-loud` is supposed to provide: not tipping off the person being scanned. It also isn't an edge case since any single module scan of a loud site hits it.

No test currently covers `__main__.py`'s handling of `-m`, so this has been silently untested.

**Suggested fix:** drop the override and pass `config` as is, matching the `-c` path. Happy to open a PR for this, could I get assigned?

A-S-Manoj:
Fixes #463

## Problem
`-m/--module` unconditionally forced `allow_loud=True`, silently overriding
the user's actual `--allow-loud` setting. Scanning a loud module via `-m`
(e.g. `user-scanner -e target@example.com -m netflix`) could send a real
notification to the target with zero warning, even without `--allow-loud`.
This was inconsistent with the `-c/--category` path, which correctly
respects it.

## Fix
Rather than just dropping the override (which would silently skip loud
modules on `-m` with no explanation), this adds a TTY-aware consent prompt,
following the existing `hudson.py` pattern:

- Loud module + no `--allow-loud` + interactive TTY → warning + `(y/n/d)` prompt
  - `y` → run once
  - `n` → skip
  - `d` → run once, save `auto_loud_single_module_prompt: false` to `config.json`
- Loud module + no `--allow-loud` + non-interactive (CI/piped, no TTY) → skip immediately with a printed warning, no hang
- `--allow-loud` passed, or saved preference set → runs without prompting, with a one-line log noting it was allowed

Skipped modules now also produce a `Result.skipped()` (same shape the
orchestrator already builds for the `-c` path),

Non-loud modules and `-c/--category` scans are unaffected.

## Note for reviewer
Loud-module prompting happens per-target inside the target loop, so a
multi-target `-m` scan (e.g. 5 targets) will prompt once per target rather
than once overall. This felt like the safer default since each target has
independent notify risk, but flagging it in case you'd prefer it prompt
once per module across all targets instead.

## Changes
- `user_scanner/core/loud_prompt.py` (new) — `check_loud_module_permission()`, `update_loud_module_preference()`
- `user_scanner/core/helpers.py` — added `auto_loud_single_module_prompt` to default config
- `user_scanner/config.json` — added the new default key (`true`)
- `user_scanner/__main__.py` — replaced blanket `allow_loud=True` override in `-m` handling with per-module consent check + proper skipped-result reporting

## Testing
Manually verified:
- `-m netflix` interactively → prompt appears, `y`/`n`/`d` behave as expected
- `-m netflix` piped/non-interactive (`< /dev/null`) → skips immediately, no hang
- `-m netflix --allow-loud` → runs directly, no prompt
- After `d`, `config.json` persists the preference and subsequent runs skip the prompt with a one-line log

No existing tests cover `-m` handling in `__main__.py`.

### DIFF
--- a/tests/test_loud_prompt.py
+++ b/tests/test_loud_prompt.py
@@ -1,0 +1,64 @@
+from user_scanner.core import loud_prompt
+from user_scanner.core.helpers import _get_config_path, load_config, save_config_value
+
+
+def test_loud_module_config_value():
+    actual_path = _get_config_path()
+    data = load_config(path=actual_path)
+    status = data.get("auto_loud_single_module_prompt")
+    assert status is True, f"FAIL: Actual config at {actual_path} has auto_loud_single_module_prompt set to {status} (Expected: True)"
+
+
+def test_check_loud_module_permission_tty_yes(monkeypatch):
+    monkeypatch.setattr(loud_prompt.sys.stdin, "isatty", lambda: True)
+    monkeypatch.setattr("builtins.input", lambda _: "y")
+    monkeypatch.setattr(loud_prompt, "load_config", lambda: {"auto_loud_single_module_prompt": True})
+    assert loud_prompt.check_loud_module_permission("Netflix", "test@example.com") is True
+
+
+def test_check_loud_module_permission_tty_no(monkeypatch):
+    monkeypatch.setattr(loud_prompt.sys.stdin, "isatty", lambda: True)
+    monkeypatch.setattr("builtins.input", lambda _: "n")
+    monkeypatch.setattr(loud_prompt, "load_config", lambda: {"auto_loud_single_module_prompt": True})
+    assert loud_prompt.check_loud_module_permission("Netflix", "test@example.com") is False
+
+
+def test_check_loud_module_permission_tty_dont_ask(monkeypatch):
+    saved = {}
+    monkeypatch.setattr(loud_prompt.sys.stdin, "isatty", lambda: True)
+    monkeypatch.setattr("builtins.input", lambda _: "d")
+    monkeypatch.setattr(loud_prompt, "load_config", lambda: {"auto_loud_single_module_prompt": True})
+    monkeypatch.setattr(loud_prompt, "update_loud_module_preference", lambda v: saved.setdefault("value", v))
+    assert loud_prompt.check_loud_module_permission("Netflix", "test@example.com") is True
+    assert saved["value"] is False
+
+
+def test_check_loud_module_permission_no_tty(monkeypatch):
+    monkeypatch.setattr(loud_prompt.sys.stdin, "isatty", lambda: False)
+    monkeypatch.setattr(loud_prompt, "load_config", lambda: {"auto_loud_single_module_prompt": True})
+    called = []
+    monkeypatch.setattr("builtins.input", lambda *_: called.append(True))
+    assert loud_prompt.check_loud_module_permission("Netflix", "test@example.com") is False
+    assert called == []  # input() never invoked
+
+
+def test_check_loud_module_permission_saved_preference(monkeypatch):
+    monkeypatch.setattr(loud_prompt, "load_config", lambda: {"auto_loud_single_module_prompt": False})
+    called = []
+    monkeypatch.setattr("builtins.input", lambda *_: called.append(True))
+    assert loud_prompt.check_loud_module_permission("Netflix", "test@example.com") is True
+    assert called == []
+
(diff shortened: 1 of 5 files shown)


kaifcodec:
@A-S-Manoj I tested possible cases and it performed as expected, but I will wait a little bit more to review it further as it touches the core files so I have to be careful.

> Thanks for the fix though and feel free to join the discord if you want to stay in touch with future issues, if you join email me your username so I could give you Contributor role there.

VamatoHD:
Firstly, there should be implemented a test in order that changes to the `config.json` aren't allowed, as the other values are checked.

Also, when running in piped/non-interactive mode, shouldn't the "user-scanner" title also be hidden, as well as the update prompt?

kaifcodec:
@VamatoHD Great catch! I agree that testing is essential here.

Before merging, I think the following additions would strengthen this PR:

**1. Update `test_helpers.py`** to assert the new config key
   - Add a test that verifies `auto_loud_single_module_prompt` defaults to `True` in the config (following the same pattern as `test_hudson_config_value()` on line 421)
   - Ensure this key doesn't get accidentally changed during config updates

**2. Add dedicated tests for `loud_prompt.py`**
   - Interactive TTY: Verify the `(y/n/d)` prompt logic and each response path
   - Non-interactive (no TTY): Confirm it skips silently with a warning instead of hanging
   - Config persistence: Verify that selecting `d` correctly saves the preference to `config.json`
   - Saved preference: Confirm that subsequent runs skip the prompt when the preference is set to `False`

---

**On @VamatoHD's second point about hiding the title/update prompt:**
I don't think that's necessary for this fix. The tool already has library mode for non-interactive use, and users deploying in CI can simply:
- Set `"auto_loud_single_module_prompt": false` in their `config.json` beforehand, OR
- Pass `--allow-loud` explicitly

This PR already prevents hanging in piped/non-interactive environments, which was the core safety concern. The title/update prompt behavior is as seperate concern and could be addressed in a follw-up PR if needed.

@A-S-Manoj if you're familiar with pytest, feel free to add these tests. Let me know if you need any guidance on the test structure!

A-S-Manoj:
Thanks @kaifcodec and @VamatoHD  for the review. I agree tests are needed here, will add coverage for the config default and the new loud_prompt module. Agree with leaving the title/update-prompt hiding out of scope for this PR, that can be a separate follow-up if needed.

Will write these and update the PR shortly.
> Converting to draft while I add the tests, will mark ready for review once they're in.

A-S-Manoj:
Added tests as requested:
- `test_loud_module_config_value` asserts the default
- `check_loud_module_permission` tests covering y/n/d, non-TTY skip, and saved-preference bypass
- `test_save_config_value_preserves_loud_module_key` confirms the key survives unrelated config updates and persists correctly when changed directly
- Ran the full test suite locally, all 148 passing

Also synced the branch with main. Marking as ready for review.

kaifcodec:
@A-S-Manoj Just a bit busy at the moment, as you can see from the recent PRs! 

From what I've reviewed so far, everything looks spot on. I'll get this merged shortly. Thanks again for your work on this, @A-S-Manoj!

kaifcodec (approved):
@A-S-Manoj LGTM! I refactored line 40 of `user_scanner/core/loud_prompt.py`. The original message was slightly misleading, targets aren't directly notified that they're being scanned, but the resulting emails serve as an indirect indicator of OSINT activity.

json-hunter07 (approved):
LGTM.

A-S-Manoj:
Thanks for the reviews and for merging this! Happy to contribute.

PR #466, opened 2026-07-26
fix: mastodon email module false positives

kristoisberg:
This PR tightens error message checks in the Mastodon email module, which was producing a false positive for every email. The previous "has already been taken" check also matched the error message for the username field. I have a suspicion that the user was actually registered by user-scanner since the module is probing the registration endpoint, in which case the module was "loud" for exactly one run.

### DIFF
--- a/user_scanner/email_scan/social/mastodon.py
+++ b/user_scanner/email_scan/social/mastodon.py
@@ -42,16 +42,22 @@ async def _check(email):
             res_text = response.text
             res_status = response.status_code
 
-            if "has already been taken" in res_text:
-                return Result.taken(url=base_url)
-            elif "registration attempt has been blocked" in res_text:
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@kristoisberg Solid catch!
Need to merge it soon, and bump the version.

kristoisberg:
> @kristoisberg Solid catch! Need to merge it soon, and bump the version.

There are a few similar issues left, which I'm fixing today, if you wish to wait for them before bumping the version. They're not as consistently annoying as the Mastodon ones and only affect less-known email providers on certain sites.

kaifcodec:
@kristoisberg No need to worry, you work on those.
I am planning to bump the version soon because a false positive is the worst error that we can have (even the false negative is better than it). 
We will bump again after those fixes.

PR #469, opened 2026-07-26
fix: flipboard email module false positives

kristoisberg:
The Flipboard email module currently reports all invalid responses as taken emails, which causes false positives with email domains, which, I assume, have been blacklisted. This PR tightens the check to only report a taken email address when Flipboard responds with error code 3109.

Previously false positive email to test with: asddfgdfdgdg@tutanota.com

### DIFF
--- a/user_scanner/email_scan/news/flipboard.py
+++ b/user_scanner/email_scan/news/flipboard.py
@@ -32,11 +32,25 @@ async def _check(email: str) -> Result:
                 return Result.error("Rate limited", url=show_url)
 
             data = response.json()
+            success = data.get("success")
+            valid = data.get("valid")
+            errorcode = data.get("errorcode")
+            reason = data.get("reason", "")
+            code = data.get("code")
 
-            if data.get("success") is True:
-                if data.get("errorcode") == 3109:
+            if success is True:
+                # 3109 means another account is already associated with this email.
+                if errorcode == 3109:
                     return Result.taken(url=show_url)
-                return Result.available(url=show_url)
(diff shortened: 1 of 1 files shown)


user_scanner/email_scan/news/flipboard.py
```diff
@@ -34,10 +34,9 @@ async def _check(email: str) -> Result:
             data = response.json()
 
             if data.get("success") is True:
-                if data.get("valid") is False:
+                if data.get("errorcode") == 3109:
                     return Result.taken(url=show_url)
-                elif data.get("valid") is True:
-                    return Result.available(url=show_url)
+                return Result.available(url=show_url)
```
kaifcodec: But here this one assumes that if it is `success` and the errorcode is not 3109 then the email is available, but that can be risky maybe for some edge cases which we hadn't yet faced, so maybe a stricter `elif` check like,
```python
elif data.get("valid") is True and data.get("errorcode" ) is None and data.get("code") == 200:
```
Anyways assign the variables for the keys first as those can be used in the `if` block for `taken()` check as well. (DRY).

kristoisberg: Done, the branches for other errors now also return the original error messages from Flipboard. I guess #470 would need similar changes since that's quite naive as well?

kaifcodec: @kristoisberg I don't think #470 needs any changes, that's perfect.
Because in there it doesn't blindly returns `Result.available()` as there it returns `.available()` status under the elif check.

kaifcodec (approved):
MERGING IT.

PR #470, opened 2026-07-26
fix: xnxx and xvideos email module false positives

kristoisberg:
The XNXX and XVideos email modules currently report all error messages as taken emails, which causes false positives with email domains, which have been blacklisted. This PR tightens the checks to only report a taken email address when  error messages specifically mention in-use emails.

Previously false positive email to test with: asdgtgffgffhg@bk.ru


kaifcodec (approved):
Perfectly done @kristoisberg !

PR #471, opened 2026-07-27
feat(export): add PDF report generator and --format pdf option

kaifcodec:
#### Description
Adds a PDF report generator feature to `user-scanner`.

#### Key Changes
- **PDF Generator Module (`user_scanner/core/pdf_generator.py`)**: Built with ReportLab, Pillow, and Svglib. Generates an Intelligence Dossier containing header details with tool version, target summary box, identified profile media grid (60x60 square cropped avatars), digital footprint mapping table, and extracted deep intelligence cards with proper vertical leading to prevent text overlap.
- **Formatter Helper (`user_scanner/core/formatter.py`)**: Added `into_pdf` wrapper function.
- **CLI Export Support (`user_scanner/__main__.py`)**: Updated `-f` / `--format` choices to accept `pdf`, auto-detected `.pdf` file extensions, and added `--no-pdf-media` flag option to skip avatar downloads if needed.
- **Optional Dependencies (`pyproject.toml`)**: Defined `pdf = ["reportlab>=4.0.0", "pillow>=10.0.0", "svglib>=1.5.0"]`.
- **Unit Tests (`tests/test_pdf_generator.py`)**: Comprehensive test suite covering PDF generation, magic byte verification, header version display, and missing dependency fallbacks.

### DIFF
--- a/user_scanner/core/pdf_generator.py
+++ b/user_scanner/core/pdf_generator.py
@@ -1,0 +1,486 @@
+import io
+import json
+from datetime import datetime
+from typing import List, Dict, Any, Union, Optional
+
+import httpx
+
+try:
+    from reportlab.lib.pagesizes import A4
+    from reportlab.lib import colors
+    from reportlab.lib.styles import getSampleStyleSheet, ParagraphStyle
+    from reportlab.platypus import SimpleDocTemplate, Paragraph, Spacer, Table, TableStyle, Image as RLImage, KeepTogether
+    from reportlab.lib.units import inch
+    REPORTLAB_AVAILABLE = True
+except ImportError:
+    REPORTLAB_AVAILABLE = False
+
+try:
+    from PIL import Image as PILImage
+    PIL_AVAILABLE = True
+except ImportError:
+    PIL_AVAILABLE = False
+
+try:
+    from svglib.svglib import svg2rlg
+    SVGLIB_AVAILABLE = True
+except ImportError:
+    SVGLIB_AVAILABLE = False
+
+
+def truncate(val: Any, max_length: int = 300) -> str:
+    if val is None:
+        return ""
+    if isinstance(val, (dict, list)):
+        try:
+            s = json.dumps(val)
+        except Exception:
+            s = str(val)
+    else:
+        s = str(val)
+        
+    s = "".join([c for c in s if ord(c) < 128])
+    if len(s) > max_length:
+        return s[:max_length] + "..."
+    return s
+
+
+def clean_metadata(extra: Any) -> List[tuple]:
+    if not extra or not isinstance(extra, dict):
+        return []
+    return [(k, v) for k, v in extra.items() if k not in ["avatar", "image"]]
+
+
+def fetch_and_resize_image(url: str, size: tuple = (60, 60)) -> Optional[Any]:
+    if not PIL_AVAILABLE:
+        return None
+    try:
+        resp = httpx.get(
+            url,
+            headers={
+                "User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/141.0.0.0 Safari/537.36"
+            },
+            timeout=3.0,
+            follow_redirects=True,
+        )
+        if resp.status_code == 200:
+            content_type = resp.headers.get("Content-Type", "")
+            if SVGLIB_AVAILABLE and ("svg" in content_type.lower() or url.lower().endswith(".svg")):
+                try:
+                    drawing = svg2rlg(io.BytesIO(resp.content))
+                    if drawing and hasattr(drawing, "width") and hasattr(drawing, "height") and drawing.width > 0 and drawing.height > 0:
+                        sx = size[0] / drawing.width
+                        sy = size[1] / drawing.height
+                        s = min(sx, sy)
+                        drawing.scale(s, s)
+                        drawing.width = drawing.width * s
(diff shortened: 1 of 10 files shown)


user_scanner/core/pdf_generator.py
```diff
@@ -0,0 +1,486 @@
...
+    )
+    elements.append(summary_table)
+
+    # Profile Photos Grid
+    if include_media and PIL_AVAILABLE:
+        unique_photos = {}
+        for idx, hit in enumerate(hits):
+            extra = hit.get("extra", {})
+            if isinstance(extra, dict):
+                url = extra.get("avatar") or extra.get("image")
```
kristoisberg: I wonder how many modules there are that store images in alternative fields. One I can think of is the Snapchat module's "snapcode" field, which technically contains a profile picture. 

I think the conventions should now be documented in CONTRIBUTING.md, or perhaps we could start storing all extracted images in another map similar to "extra", in case some sites provide multiple images (e.g. YouTube profile picture and banner). There, each value would be guaranteed to be an URL to an image, so logic like this wouldn't need key-name heuristics.

kristoisberg:
When testing the snapcode question I noticed some issues with link rendering:
`user-scanner -f pdf -u asdfg -m snapchat`

>  Checking username: asdfg
>   [✔] Snapchat (asdfg): Found
>       └── snapcode: https://app.snapchat.com/web/deeplink/snapcode?username=asdfg&type=SVG&bitmoji=enable

But it is rendered in the report as follows:
image

Only the first line is clickable, and some characters are added that shouldn't be there.

Looks great otherwise!

user_scanner/core/pdf_generator.py
kaifcodec: @kristoisberg Yeah, that is spot on! I noticed a few sites missed the pfp image URL simply because the key name was `avatar_url`.

Your approach of introducing another separate parameter similar to `extra` sounds like a solid path. Though in that case, we’d have to refactor existing modules to pass `avatar` or `image_url` properly. If you have a plan ready, feel free to open an issue about it (doesn't need to be super detailed, just enough to map out the direction). I really like the idea, if we ever decide to expand into deeper scans like extracting target posts, it'll definitely be a lifesaver feature.

Also, I'll take care of the issue with unexpected snapcode URL modifications during PDF generation. I noticed another minor logic error as well: it currently prints "[+] Results saved to ..." in green even when no new file was actually created (when no `-f` flag is passed).

kaifcodec:
@kristoisberg The issue happened because ReportLab’s `Paragraph` component parses text input as HTML/XML. When a URL contains unescaped ampersands (`&`), ReportLab mistakes them for XML entities and automatically injects semicolons, turning `&type` into `&type;=` and breaking the link.

Wrapping the raw URL strings with `html.escape()` before passing them to `Paragraph` ensures `&` is properly handled as `&amp;`, keeping the URL intact without any unwanted semicolons.

I pushed the fix in (31e9326).

user_scanner/core/pdf_generator.py
```diff
@@ -0,0 +1,487 @@
...
+            ]
+        )
+    )
+    elements.append(header_table)
+    elements.append(Spacer(1, 20))
+
+    hits = [
+        r
+        for r in raw_results
+        if str(r.get("status", "")).lower() in ["found", "registered", "taken"]
```
VamatoHD, line 231: When looking at `result.py`
``` 
def to_label(self, is_email=False):
    if self == Status.ERROR:
        return "Error"
    elif self == Status.SKIPPED:
        return "Skipped"
    if is_email:
        return "Registered" if self == Status.TAKEN else "Not Registered"
    return "Found" if self == Status.TAKEN else "Not Found"
``` 
This check ends up ignoring `Status.Error`, `Status.Skipped` and `Status.available`, so the PDF table that is generated only shows the modules that led into a taken username. Therefore, that column is kind of unnecessary, as every single one just says "Found".

user_scanner/core/pdf_generator.py
```diff
@@ -0,0 +1,487 @@
...
+            img = img_raw.convert("RGB")
+
+            # Crop to square
+            min_dim = min(img.width, img.height)
+            left = (img.width - min_dim) / 2
+            top = (img.height - min_dim) / 2
+            right = (img.width + min_dim) / 2
+            bottom = (img.height + min_dim) / 2
+            cropped = img.crop((left, top, right, bottom))
+            resized = cropped.resize(size, PILImage.Resampling.LANCZOS)
```
VamatoHD: @kaifcodec is there any reason for scaling the profile pictures into 60x60? There could be a limit instead (of around 600x600) where every picture would be scaled down if it was bigger than the limit. Further into the PDF generation, the element that holds the images could be 60x60 but holds much higher quality images.

user_scanner/core/pdf_generator.py
kaifcodec: @VamatoHD Yeah, you can say that but I kept it like that just to fill it out and thinking if we ever want to implement something where we have to show status, about something else (maybe a new flag later that will keep the `Not Found`s as well) it will work in that time.

Even I am not sure what to do with that, I was working on this last month so kept that as it was there.

Does it look bad?

user_scanner/core/pdf_generator.py
kaifcodec: Yeah I agree on this, I updated the max_size to (600x600) in the last commit (2f519f6).

kristoisberg (approved):

VamatoHD (approved):
LGTM

PR #474, opened 2026-07-27
fix: BeSoccer email module false positives

kristoisberg:
Another case of a module depending on an error code that's too broad, causing false positives with blacklisted email domains.

Test case: any mail.ru email address

I also slightly reduced the length of the generated usernames as the username length limit is 20, but the code generated 21-character usernames, causing additional noise.


kaifcodec (approved):
@kristoisberg LGTM.

PR #475, opened 2026-07-27
fix: YouTube username module false positives

kristoisberg:
This PR fixes the YouTube username module reporting false positives with all usernames. The module currently reports the username as taken whenever the visited page does not explicitly say that the user does not exist. However, at least in my case, the request does not even reach the actual target as it is redirected to a YouTube consent page, and the module considers this valid.

1. adds `?cbrd=1&ucbcb=1` query parameters, which skip the YouTube consent page and go directly to the user profile or 404 page
2. only reports the username as taken if there are clear indications of it; reports an error in case of an ambiguous state

I think I have patched all false positives I've discovered now.

### DIFF
--- a/user_scanner/user_scan/social/youtube.py
+++ b/user_scanner/user_scan/social/youtube.py
@@ -1,10 +1,13 @@
-from user_scanner.core.orchestrator import Result, make_request
-import re
 import json
+import re
+
+import httpx
+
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@kristoisberg It's perfect and the issue is fixed, by the way really impressive, how detailed you check the issues.

PR #481, opened 2026-07-29
feat: add Nexus Mods username module

kristoisberg:
This PR adds an username module for the gaming site nexusmods.com. One of the simpler modules yet, since the site exposes E2E testing attributes for nearly everything of interest.

### DIFF
--- a/user_scanner/user_scan/gaming/nexusmods.py
+++ b/user_scanner/user_scan/gaming/nexusmods.py
@@ -1,0 +1,48 @@
+import html
+import re
+from urllib.parse import quote
+
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
+
+
+def validate_nexusmods(user: str) -> Result:
+    url = f"https://www.nexusmods.com/profile/{quote(user, safe='')}"
+
+    def process(response):
+        if response.status_code == 404:
+            return Result.available()
+
+        if response.status_code != 200:
+            return Result.error(f"Unexpected status code: {response.status_code}")
+
+        heading = re.search(
+            r'<h1[^>]*class="[^"]*text-heading-sm[^"]*"[^>]*>\s*([^<]+)\s*</h1>',
+            response.text,
+        )
+        if not heading or html.unescape(heading.group(1)).strip().casefold() != user.casefold():
+            return Result.error("Profile response did not match the requested username")
+
+        extra = {}
+        avatar = re.search(
+            r'src="(https://avatars\.nexusmods\.com/[^"]+)"', response.text
+        )
+        if avatar:
+            extra["avatar"] = html.unescape(avatar.group(1))
+
+        for field, marker in (
+            ("endorsements_given", "endorsements-given"),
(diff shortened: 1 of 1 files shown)


user_scanner/user_scan/gaming/nexusmods.py
```diff
@@ -0,0 +1,48 @@
...
+
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
+
+
+def validate_nexusmods(user: str) -> Result:
+    url = f"https://www.nexusmods.com/profile/{quote(user, safe='')}"
+
+    def process(response):
+        if response.status_code == 404:
```
kaifcodec: @kristoisberg I see it currently relies solely on a `404` status code to check if a username is available. That works for now, but there's a slight risk, if they ever change or deprecate the API endpoint, a missing route will return a `404` and start triggering false negatives for every user.

If you can find a unique response body marker (like a specific error message or key) that only appears when a username is actually available/not found, could you add that check to the `if` condition with an `and` operator?

kristoisberg: Done, there was a marker indicating that the profile page route was matched.

brunolm (commented):
1. `generic_validate` (httpx) is Cloudflare-blocked — every request returns `403` (`Just a moment...`), so the module always errors and never returns taken/available. Live check: same URL via `impersonate_request(..., allow_redirects=True)` returns `200` for real profiles (`Gopher`, `Pickysaurus`, `Dark0ne`) and `404` for fakes. Switch to `impersonate_validate` and use `allow_redirects=True` (curl_cffi kwarg, not `follow_redirects`). Homepage `warmup_url` is optional; fingerprint alone was enough here.
2. `joined` / `last_active` scrape display text (`07 Sept 2004`) while the same `<time>` elements expose ISO `dateTime` (e.g. `2004-09-07T10:55:16.000Z`). Prefer the `dateTime` attribute for locale-independent, machine-sortable values.
3. Also extract `data-e2eid="unique-download-count"` (present on live profiles, e.g. `9.4m`) into `extra`, same pattern as the other e2e fields.
4. Avatar URLs are `https://avatars.nexusmods.com/<user_id>/100` — optionally emit the numeric id into `extra` (and/or prefer `og:image` for the avatar).


```
Raw output:

 Checking username: brunolm
  [!] Nexusmods [https://www.nexusmods.com/profile/brunolm] (brunolm): Error (Unexpected status code: 403)


[i] Scan complete.
  Total hits: 0
```
Matches the PR review: httpx hits Cloudflare (403), so the module never gets a profile response. No JSON file was written (nothing to save with only an error and no -o).

kristoisberg:
@brunolm Thank you for the suggestions, I applied them all. I also parsed some additional information from the page's hydration data.

brunolm (approved):

kaifcodec (approved):
@kristoisberg This is really a high quality module as it reveals the targets' email.
@brunolm Thanks for review on this.

MERGING IT!

PR #482, opened 2026-07-29
feat: add Polarsteps username module

kristoisberg:
Username module for the travel tracking app polarsteps.com. The enhancements are very detailed here since we could access them directly from their API with little effort.

### DIFF
--- a/user_scanner/user_scan/other/polarsteps.py
+++ b/user_scanner/user_scan/other/polarsteps.py
@@ -10,7 +10,7 @@ def validate_polarsteps(user: str) -> Result:
     show_url = f"https://www.polarsteps.com/{encoded_user}"
     headers = {
         "Accept": "application/json",
-        "Polarsteps-API-Version": "61",
+        "Polarsteps-API-Version": "61",  # bump if module starts erroring on all lookups
     }
 
     def process(response):
@@ -21,6 +21,12 @@ def validate_polarsteps(user: str) -> Result:
 
         if not isinstance(data, dict):
             return Result.error("Unexpected response body")
+
+        if data.get("error") == "api_version_too_old_for_endpoint":
+            return Result.error(
+                f"Polarsteps API version outdated — requires v{data.get('required_version')}, "
+                f"update 'Polarsteps-API-Version' header in polarsteps.py"
+            )
 
         if response.status_code == 404 and data == {"detail": "Not Found"}:
             return Result.available()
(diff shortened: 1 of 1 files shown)


user_scanner/user_scan/other/polarsteps.py
```diff
@@ -0,0 +1,70 @@
...
+from user_scanner.core.result import Result
+
+
+def validate_polarsteps(user: str) -> Result:
+    encoded_user = quote(user, safe="")
+    api_url = f"https://api.polarsteps.com/users/byusername/{encoded_user}"
+    show_url = f"https://www.polarsteps.com/{encoded_user}"
+    headers = {
+        "Accept": "application/json",
+        "Polarsteps-API-Version": "61",
```
A-S-Manoj: Nit: this hardcoded version is a bit fragile long-term — Polarsteps already deprecated the legacy `users/byusername` endpoint once before for anyone below v52.

If this module ever starts returning `error` for every lookup, this is the first thing to check.

```suggestion
        "Polarsteps-API-Version": "61",  # bump if module starts erroring on all lookups
```

user_scanner/user_scan/other/polarsteps.py
```diff
@@ -0,0 +1,70 @@
...
+    def process(response):
+        try:
+            data = response.json()
+        except ValueError:
+            return Result.error(f"Unexpected non-JSON response: {response.status_code}")
+
+        if not isinstance(data, dict):
+            return Result.error("Unexpected response body")
+
+        if response.status_code == 404 and data == {"detail": "Not Found"}:
```
A-S-Manoj, line 31: Suggestion: worth explicitly catching the API-version-gate error before the 404 check, so a future version bump fails with a clear, actionable message instead of falling through to a generic "unexpected response status" error. Not blocking — the code already fails safe either way.

```suggestion
        if data.get("error") == "api_version_too_old_for_endpoint":
            return Result.error(
                f"Polarsteps API version outdated — requires v{data.get('required_version')}, "
                f"update 'Polarsteps-API-Version' header in polarsteps.py"
            )

        if response.status_code == 404 and data == {"detail": "Not Found"}:
```

A-S-Manoj (commented):
Nice work on this one — logic looks solid and follows the existing validator pattern well. Left a couple of non-blocking suggestions inline (mostly around long-term robustness of the hardcoded API version), but nothing that should hold up the merge. LGTM overall

kristoisberg:
@A-S-Manoj Thanks for the review and suggestions, failing loudly is a great idea.

kaifcodec (approved):
@kristoisberg This is perfect now and working.
Thank you @A-S-Manoj for the review on this.

MERGING IT!

PR #486, opened 2026-08-01
fix: skip update prompt when PyPI check fails (#480)

### LINKED ISSUE #480: [Bug] Failed PyPI version check is shown as a false "new version available" prompt
**Summary**

When the PyPI version check fails (offline, PyPI down, rate limited, etc.), the tool incorrectly reports that a new version is available.

**Where**

`user_scanner/utils/updater_logic.py`, `check_for_updates()`

**Details**

`get_pypi_version()` catches any request failure and returns `None`:

```python
def get_pypi_version(pypi_url):
    try:
        pypi_version = httpx.get(pypi_url, timeout=7).json()["info"]["version"]
    except Exception as e:
        print(e)
        return None
    return pypi_version
```

`check_for_updates()` compares this directly without checking for `None`:

```python
if current_ver != latest_ver:
    print(f"\n[!] New version available: {R}{current_ver}{X} -> {C}{latest_ver}{X}\n")
    choice = input("Do you want to update? (y/n/d): ")
```

Since `current_ver` is always a real string, this is `True` on every failed check, producing:

[!] New version available: 1.4.1.9 -> None

Reproduced by mocking `get_pypi_version` to return `None` and calling `check_for_updates()` directly.

**Impact**

Fires on any network failure, not an edge case. Also feeds into `update_self()`, which uninstalls before reinstalling, so accepting the prompt while offline risks leaving the tool uninstalled.

**Suggested fix**

Skip the prompt when `latest_ver is None`.

I'd like to work on this myself.

A-S-Manoj:
**Fixes #480**

## Problem

`get_pypi_version()` catches any request failure (offline, PyPI down, rate limited, etc.) and returns `None` instead of raising. `check_for_updates()` compared this directly against the local version with no `None` check, so `current_ver != None` was always true on any failed check, producing a false prompt:

[!] New version available: 1.4.1.9 -> None
Do you want to update? (y/n/d: don't ask again):

This isn't a rare edge case, it fires on any network failure. It also feeds into `update_self()`, which uninstalls before reinstalling, so accepting the prompt while genuinely offline risks leaving the tool uninstalled.

## Fix

Added a check for `latest_ver is None` immediately after it's fetched, and return early with a clear message instead of comparing it against `current_ver`:

- Could not reach PyPI → print `"[!] Could not reach PyPI, skipping update check."` and return, no prompt shown
- Real version mismatch → unchanged, still prompts as before
- Versions match → unchanged, stays silent

## Changes

- `user_scanner/utils/updater_logic.py` — added the `None` check in `check_for_updates()` before the version comparison
- `tests/test_updater_logic.py` (new) — regression coverage for the three cases above

## Testing

Ran the full test suite against current `main` (151 passed). Verified the new tests fail on the unmodified code and pass with the fix:

- PyPI unreachable → no prompt shown, `input()` never called
- Real update available → prompt still shown
- Versions match → no prompt, `input()` never called

Without the fix, the unreachable case doesn't just show a false prompt, it goes on to crash with `'NoneType' object has no attribute 'strip'` once `input()` is unexpectedly reached, since the mocked/aborted input path returns `None` instead of a string.

No existing test covered this path. The closest one, `test_pypi_error_is_caught` in the incoming `test_updater_logic.py` PR, mocks `get_pypi_version` to raise rather than return `None`, so it doesn't exercise this bug.

### DIFF
--- a/tests/test_updater_logic.py
+++ b/tests/test_updater_logic.py
@@ -1,0 +1,56 @@
+from user_scanner.utils import updater_logic
+
+
+def test_skips_prompt_when_pypi_unreachable(monkeypatch, capsys):
+    monkeypatch.setattr(
+        updater_logic, "load_config", lambda: {"auto_update_status": True}
+    )
+    monkeypatch.setattr(updater_logic, "get_pypi_version", lambda url: None)
+    monkeypatch.setattr(
+        updater_logic, "load_local_version", lambda: ("1.4.1.9", "local")
+    )
+
+    called = []
+    monkeypatch.setattr("builtins.input", lambda *_: called.append(True))
+
+    updater_logic.check_for_updates()
+
+    out = capsys.readouterr().out
+    assert "New version available" not in out
+    assert "Could not reach PyPI" in out
+    assert called == []  # input() never invoked
+
+
+def test_prompts_when_real_update_available(monkeypatch, capsys):
+    monkeypatch.setattr(
+        updater_logic, "load_config", lambda: {"auto_update_status": True}
+    )
+    monkeypatch.setattr(updater_logic, "get_pypi_version", lambda url: "9.9.9")
+    monkeypatch.setattr(
+        updater_logic, "load_local_version", lambda: ("1.4.1.9", "local")
+    )
+    monkeypatch.setattr("builtins.input", lambda _: "n")
+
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
@A-S-Manoj LGTM.
Thank you for the tests.

PR #487, opened 2026-08-01
fix: reject usernames causing false positives in academia.edu module

kristoisberg:
Similar to #437, rejecting usernames with periods in the academia.edu module. The site does not allow usernames with periods, and ignores everything after the first period in profile URLs, causing false positives in the module.

Test case: https://independent.academia.edu/asdfg.fghfghhfhgfhgfghfghghfgh


kaifcodec (approved):
LGTM.

PR #488, opened 2026-08-01
fix: Bluesky username module false positives

kristoisberg:
This PR fixes false positives in the Bluesky username module. In addition to taken usernames, the currently used "checkHandleAvailability" RPC call also rejects usernames that are invalid for other reasons.

I removed the "checkHandleAvailability" RPC call, instead checking username availability using the "getProfile" RPC, which was already used to fetch enrichments. As a result, the module also became much simpler.

Previously false positive test case:
`user-scanner -u dgfghhfghfgfghijogh --verbose -m bluesky`


kaifcodec (approved):
@kristoisberg I just refactored the error message as this one tells the user to report it.

json-hunter07 (approved):
@kaifcodec LGTM.

PR #489, opened 2026-08-01
feat: add Dropbox email module

beranka:
Adds an email-existence scan module for Dropbox ([its home page](https://www.dropbox.com/)).

## How it works

- Retrieves the CSRF token from cookies set when requesting the main page.
- Uses Dropbox's endpoint that explicitly returns whether an email is registered.
- The endpoint also returns whether the email is used as a recovery email for another Dropbox account and whether the associated account has a passkey configured.

## Testing

Manually tested (and got expected results) with:

- An email that is not registered.
- An email that is only registered as a recovery email.
- An email that is registered and the associated account does not have a passkey.
- An email that is registered and the associated account has a passkey.

### DIFF
--- a/user_scanner/email_scan/other/dropbox.py
+++ b/user_scanner/email_scan/other/dropbox.py
@@ -1,0 +1,88 @@
+import contextlib
+
+import httpx
+
+from user_scanner.core.result import Result
+
+
+async def _check(email: str) -> Result:
+    """
+    Checks whether a given email is registered on Dropbox.
+
+    The check has two steps:
+
+    1. Retrieve CSRF token (is set as a cookie with name "t").
+    2. Use the check_user_with_email_exists endpoint of Dropbox's API to retreive the result.
+
+    The API also returns whether the given email is registered as a recovery email and whether it has a passkey.
+    """
+
+    show_url = "https://www.dropbox.com"
+    check_user_url = f"{show_url}/2/account/check_user_with_email_exists"
+
+    headers = {
+        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:150.0) Gecko/20100101 Firefox/150.0",
+        "Referer": show_url,
+        "Origin": show_url,
+        "Accept": "application/json",
+        "Content-Type": "application/json",
+    }
+
+    async with httpx.AsyncClient(timeout=15.0, follow_redirects=True) as client:
+        # 1. Retrieve the CSRF token
+        csrf_token_retrieval_response = await client.get(show_url, headers=headers)
+
+        if csrf_token_retrieval_response.status_code != 200:
+            return Result.error(
+                f"Failed to access the {show_url}: {csrf_token_retrieval_response.status_code}"
+            )
+
+        # Dropbox sets the CSRF token as the "t" cookie. The same value is also mirrored in "__Host-js_csrf".
+        csrf_token = client.cookies.get("t")
+        if not csrf_token:
+            return Result.error("CSRF token not retrieved")
+
+        # 2. Retrieve the information from Dropbox's API
+        headers["X-CSRF-Token"] = csrf_token
+        payload = {"email": email}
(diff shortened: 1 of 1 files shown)


user_scanner/email_scan/other/dropbox.py
```diff
@@ -0,0 +1,88 @@
...
+    has_passkey = data.get("has_passkey")
+
+    for value in (exists, exists_as_recovery_email, has_passkey):
+        if not isinstance(value, bool):
+            return Result.error(
+                "Failed to check, the response structure may have changed. Please create Github issue."
+            )
+
+    extra = {
+        "exists": exists,
```
kaifcodec: @beranka There's no need to pass the `exists` boolean via the `extra` dict. The "Registered" / "Not Registered" status already handles and conveys that distinction.

I tested it locally and I didn't find any issue on the first run, just needed this fix for now.

beranka: I return the Registered (or taken, or whatever should be the terminology) result even when the email is only used as a recovery email (and not actually as the main email).

I think it makes sense to flag it as found in that case, and therefore there needs to be a way to distinguish between "not registered but is recovery for another account", "registered and also recovery for another account". This way of distinguishing it felt most natural to me.

Let me know if you think this should be done differently! Thanks :)

kaifcodec: @beranka I see where you're coming from, but take a look at lines 82–84.

When either `exists` is `True` **or** `exists_as_recovery_email` is `True`, the module correctly returns `Result.taken(...)`, so the overall result is already handled correctly.

The issue is that you're also passing the entire `extra` dictionary to `Result.available(...)` when **both** of those values are `False`. In that case, `extra` will always contain only negative values, and we generally don't attach `extra` to `Result.available(...)` because there isn't any additional metadata to expose when the email isn't associated with an account.

Regarding my previous comment, I wasn't referring to `exists_as_recovery_email` or `has_passkey`. I only meant that the `exists` boolean itself doesn't need to be included in `extra`, because whether the email exists is already represented by returning `Result.taken(...)` or `Result.available(...)`.

So I'd suggest keeping only the additional metadata in `extra`, for example:

```python
extra = {
    "exists_as_recovery_email": exists_as_recovery_email,
    "has_passkey": has_passkey,
}
```

This still preserves the distinction you're talking about, for example, whether the email is used as a recovery email or has a passkey, without duplicating the registration status that `Result` already conveys.

beranka: I just tried adding an email that is primary for one account as recovery for another one and I got this error:
```json
{
  "error":{".tag":"email_is_primary_of_another_user"},
  "error_summary":"email_is_primary_of_another_user/...",
  "user_message":{"locale":"en","text":"Email is not available"}
}
```
Meaning that it probably shouldn't happen for one email to be both recovery and primary one, meaning that the cases I talked about do not have to be distinguished as one of them is impossible.

So for the `exists = True` case, I will return `exists_as_recovery_email` and `has_passkey`, for `exists_as_recovery_email = True`, I will return only `exists_as_recovery_email`.

And for the case when neither is true, I will return no extras, as they will be always negative, like you said.

Done in https://github.com/kaifcodec/user-scanner/pull/489/commits/49b8f2634bf378569ad38d907566873d466464f9

json-hunter07 (approved):
@kaifcodec  LGTM.
The last commit fixed the logic error you were talking about.

kristoisberg:
Looks solid, LGTM too!

kristoisberg (approved):

kaifcodec:
@beranka Thank you for the PR, this one really is a solid addition as it reveals some `extra` data from emails. Feel free to add more new sites to the project.
**MERGING IT!**

@json-hunter07 @kristoisberg Thank you for the review on this.

tiehardoim-coder (commented):
FaEr

PR #490, opened 2026-08-02
fix(facebook): update stale unavailable marker string

### LINKED ISSUE #484: [Bug / Refactor] Fix non-existent profile detection in facebook.py
## Description
The `facebook.py` module is failing when checking non-existent/available usernames, returning `Error (Unexpected response body, report it via GitHub issues.)` instead of `Result.available()`. 

Facebook has likely updated its HTML response body or meta tags for missing profiles, meaning the hardcoded `UNAVAILABLE_MARKER` (`"This content isn't available right now"`) no longer matches.

## Technical Details
* **Module Path:** `user_scanner/user_scan/social/facebook.py`
* **Affected Function:** `_process_profile_response()`
* **Current Error:** `Error (Unexpected response body, report it via GitHub issues.)`

## Instructions for Contributors
1. **Analyze the Response:** Fetch the HTML response for a non-existent Facebook username using `curl`, DevTools, or a proxy.
2. **Find New Marker:** Inspect the response body to identify the updated text string or page title Facebook now uses for missing/unavailable profiles.
3. **Update Validation Logic:** Update `UNAVAILABLE_MARKER` or the `if` block check in `_process_profile_response()` so missing profiles successfully return `Result.available()`.
4. **Local Verification:** Run the tool locally (`user_scanner -u <username> -m facebook`) on both valid and available usernames to verify the fix works cleanly without throwing unexpected response body errors.

A-S-Manoj:
## Summary

Facebook updated the copy on its profile-not-found interstitial from **"This content isn't available right now"** to **"This content isn't available at the moment"**. The hardcoded `UNAVAILABLE_MARKER` in `facebook.py` no longer matched this string, so `_process_profile_response()` fell through to the generic catch-all and returned `Result.error("Unexpected response body, report it via GitHub issues.")` instead of `Result.available()` for non-existent usernames.

Fixes #484

## Root Cause

Diffed raw HTML (using full browser-style request headers to get past Facebook's bot detection) between a known-existing profile and a non-existent username:

- The `og:title` / `og:url` meta tag based "taken" detection still works correctly — no change needed there.
- Only the "not found" marker string had drifted from Facebook's side.

## Fix

```diff
- UNAVAILABLE_MARKER = "This content isn't available right now"
+ UNAVAILABLE_MARKER = "This content isn't available at the moment"
```

## Testing

Ran locally against a mix of known-taken and known-available usernames:

### DIFF
--- a/user_scanner/user_scan/social/facebook.py
+++ b/user_scanner/user_scan/social/facebook.py
@@ -8,8 +8,7 @@ PROFILE_TITLE_RE = re.compile(r'<meta property="og:title" content="[^"]+"')
 PROFILE_URL_RE = re.compile(
     r'<meta property="og:url" content="https://www\.facebook\.com/[^"]+"'
 )
-UNAVAILABLE_MARKER = "This content isn't available right now"
-
+UNAVAILABLE_MARKER = "This content isn't available at the moment"
 
 def _process_profile_response(status_code: int, html: str) -> Result:
     if status_code == 429:


kaifcodec (approved):
@A-S-Manoj Thank you for the fix.
I tested it and it works now.

PR #492, opened 2026-08-02
feat: add email module for Alza

beranka:
Adds an email scan module for Alza.
- [English Wikipedia article about the company](https://en.wikipedia.org/wiki/Alza.cz)
- [English version of its Hungarian website](https://www.alza.hu/EN/)

Alza has multiple websites for different countries in Central Europe; this checks all of them in a loop.

## How it works
For each website:
- Retrieves cookies when requesting the main page.
- Uses an API endpoint which returns:
  - `LoginAvailabilityType` field, set to:
    - `1` (account associated with the email),
    - `2` (email used for an order without an account),
    - `0` (email not found),
    - there might be other values, but those did not appear when testing different emails.
  - Other fields are also returned, but I did not notice any value changes during testing.

Then returns a `Result` with status:
- `available` if all websites returned `LoginAvailabilityType: 0`.
- `taken` if at least one website returned `LoginAvailabilityType: 1` or `LoginAvailabilityType: 2`; also returns two comma-separated lists:
  - websites where the email is associated with an account,
  - websites where the email was used for an order without an account.
- `error` if an error occurred while checking any of the websites.

## Testing

Manually tested (and got expected results) with:

- Email belonging to only one account on one website (for each website)
- Email registered nowhere and not used for any orders
- Email belonging to an account on each website
- Email used previously for an order without an account on one of the websites

---

It's a little bit more complex than I expected (sorry), but I think it makes sense to check each of the regional websites :slightly_smiling_face:

### DIFF
--- a/user_scanner/email_scan/shopping/alza.py
+++ b/user_scanner/email_scan/shopping/alza.py
@@ -1,0 +1,134 @@
+import httpx
+
+from user_scanner.core.result import Result, Status
+
+ALZA_COUNTRY_CODES = ("cz", "sk", "de", "hu", "at")
+
+
+async def _check_given_website(email: str, main_url: str) -> Result:
+    """
+    Checks whether a given email is associated with an account or order on a specified Alza regional website.
+
+    The check has two steps:
+
+    1. Retrieve cookies by fetching the main page.
+    2. Use the CheckLoginAvailability API endpoint. Example response:
+        {
+            "LoginAvailabilityType": 1,
+            "DevErrorMessage": null,
+            "Message": null,
+            "ErrorNeoPurchaseFailed": false,
+            "ErrorLevel": 0,
+            "RedirectUrlOrderDetail": null,
+            "PaymentAction": null,
+            "CanShowFastCheckoutButton": false
+        }
+        The value of LoginAvailabilityType indicates whether the email is associated with a registered account (1),
+        was used for an order but is not associated with an account (2), or neither (0).
+
+        I do not know whether the other fields can have different values; they remained unchanged during testing.
+    """
+
+    check_login_availability_url = (
+        f"{main_url}/Services/EShopService.svc/CheckLoginAvailability"
+    )
+
+    headers = {
+        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:150.0) Gecko/20100101 Firefox/150.0",
+        "Referer": main_url,
+        "Origin": main_url,
+        "Accept": "application/json",
+    }
+
+    async with httpx.AsyncClient(timeout=15.0, follow_redirects=True) as client:
+        # Fetch the main page to retrieve cookies
+        response = await client.get(main_url, headers=headers)
+
+        if response.status_code != 200:
+            return Result.error(
+                f"Failed to access the {main_url}, HTTP {response.status_code}"
+            )
+
+        # Send post request to the API
+        payload = {"login": email}
+        response = await client.post(
+            check_login_availability_url, headers=headers, json=payload
+        )
+
+    if response.status_code != 200:
+        return Result.error(
+            f"Failed to retrieve information from API, HTTP {response.status_code}"
+        )
+
+    try:
+        data = response.json()
+        login_availability_type = data.get("LoginAvailabilityType")
+    except (ValueError, AttributeError):
+        return Result.error(
+            "Unexpected response structure, please report it via GitHub issues"
+        )
+
+    match login_availability_type:
+        case 0:
+            # Not registered, no orders with this email
+            return Result.available(url=main_url)
+        case 1:
+            # Account with this email exists
+            return Result.taken(url=main_url, extra={"has_account": True})
+        case 2:
+            # Orders with this email were created, but the email does not belong to any account
+            return Result.taken(url=main_url, extra={"has_account": False})
+        case _:
+            return Result.error(
+                "Unexpected response structure, please report it via GitHub issues"
+            )
+
+
+async def _check(email: str) -> Result:
+    """
+    Check whether an email is associated with an account or order across the following Alza regional websites:
+
+        - Czech Republic (www.alza.cz)
+        - Slovakia (www.alza.sk)
+        - Germany (www.alza.de)
+        - Hungary (www.alza.hu)
+        - Austria (www.alza.at)
+
+    Returns:
+        Result:
+            - with status=TAKEN if email was found on at least one website, also returns the list of websites where the given email
+                is associated with an account and a list of websites where the email was used for an order without a registered account,
+            - with status=AVAILABLE if the email was not found on any of the websites,
+            - with status=ERROR if an error occurred while checking any of the websites.
+    """
+    domain_name_template = "https://www.alza.{cctld}"
+    order_only_countries = []
+    account_countries = []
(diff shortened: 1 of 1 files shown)


kristoisberg:
@kaifcodec Thoughts on this? When I added the two regional versions of Yaga, I created separate modules for each. Here, with five sites, this would get a bit spammy...

user_scanner/email_scan/shopping/alza.py
```diff
@@ -0,0 +1,134 @@
...
+    match login_availability_type:
+        case 0:
+            # Not registered, no orders with this email
+            return Result.available(url=main_url)
+        case 1:
+            # Account with this email exists
+            return Result.taken(url=main_url, extra={"has_account": True})
+        case 2:
+            # Orders with this email were created, but the email does not belong to any account
+            return Result.taken(url=main_url, extra={"has_account": False})
```
kristoisberg, line 80: This case will definitely become confusing for users without reading source code - the module will report the email as "registered", but simultaneously says an account does not exist. I'm not even sure how to express these results better... I definitely wouldn't exclude this case, since even "has made an order" is useful information.

kaifcodec: @beranka

1. You don't need to pass the `extra` dictionary for **case 1**. Returning `Result.taken(...)` already conveys that `has_account=True`, so including it in `extra` is redundant.

2. For **case 2**, you can use the `reason` parameter instead. All three result types (`.taken()`, `.error()`, and `.available()`) support it, so you can do something like:

```python
return Result.taken(
    url=main_url,
    reason="This email is not registered, but it has been used to place an order.",
    extra={"has_account": False},
)
```
@kristoisberg
I think this better represents the OSINT value of the result. Even if the email isn't registered, it has still interacted with the site, so returning `Result.taken(...)` is reasonable. The `reason` explains why it's considered a hit, while `extra` provides the additional context that the email doesn't actually have an account.

kaifcodec:
> @kaifcodec Thoughts on this? When I added the two regional versions of Yaga, I created separate modules for each. Here, with five sites, this would get a bit spammy...

@kristoisberg Good catch. They most likely use separate databases for each regional site rather than treating all five domains as a single service.

So @beranka, I think it's better to open another PR where you split these into five separate modules. It will be more scalable and also more user-friendly, since users can immediately identify the region from the results instead of only getting to know that the email is registered on "Alza".

The filenames could be something like:

- `alza_cz.py`
- `alza_hu.py`
- `alza_sk.py`
- `alza_at.py`
- `alza_de.py`

> Our helper functions already convert `_` to `.` when importing modules, and `.` back to `_` when parsing `args.module`, so you don't need to worry about that part. Just use underscores in the filenames.

kaifcodec:
@beranka Would you like to take more time working on this PR, or should I label it as `help wanted` so other contributors can take this and work on it?

beranka:
> @beranka Would you like to take more time working on this PR, or should I label it as `help wanted` so other contributors can take this and work on it?

I am just busy with work this week, but I will take a look at it on Sunday.

kaifcodec:
That's great. Thank you for the clarification.

beranka:
> So @beranka, I think it's better to open another PR where you split these into five separate modules. It will be more scalable and also more user-friendly, since users can immediately identify the region from the results instead of only getting to know that the email is registered on "Alza".

@kaifcodec Why another PR?

Also what should I do to avoid repeating the same code for each variant? Could I maybe create `alza_common.py` somewhere? Or keep the `_check_given_website()` in one of the variants and import it to others? I don't like the idea of having 5 almost identical files...

kaifcodec:
1. **PR Workflow:** You can keep using this current PR and push the refactored commit here but a new PR is recommended for keeping the commit history clean.

2. Don't worry about minor header/URL string overlap across the 5 files. We explicitly prefer having 5 straightforward, independent module files (`alza_cz.py`, `alza_sk.py`, etc.) over creating complex helper abstractions or cross-importing between modules. 

   Keeping each module self-contained makes it super easy to maintain, if Alza changes their `.hu` backend or cookie requirements tomorrow, anyone can update `alza_hu.py` directly without digging through shared helper abstractions or risking breaking `.cz`, also you can't create some helper like `_common.py` as it goes against contributing guidelines and standard user-scanner module import patterns.

Azaucifer:
@beranka Hi! 👋

Just checking in on this PR. Are you still working on splitting this into 5 separate modules as requested?

If you're busy and don't have time, I'd be happy to take it over and finish it up.

Let me know! 🙌

beranka:
> @beranka Hi! 👋
> 
> Just checking in on this PR. Are you still working on splitting this into 5 separate modules as requested?
> 
> If you're busy and don't have time, I'd be happy to take it over and finish it up.
> 
> Let me know! 🙌

Hi, I have kind of forgotten about this, being busy with other things 😅

Feel free to take over 😉

Azaucifer:
Oh I can understand, not a problem!  
Thank you @beranka :)

Azaucifer:
## Update on Alza Module

@kaifcodec I investigated the Alza module. Here's what I found:

### Issue

Alza's API endpoint now returns 403 for all automated requests.

### What I Tried

- `httpx` with headers → ❌ 403
- `curl_cffi` with Chrome impersonation → ❌ 403
- Playwright (page loads) → ✅ 200
- Playwright (API calls) → ❌ 403

### Conclusion

The Alza API is now protected and blocks all automated requests, even through a real browser context.

**Recommendation:** This module can be moved to `abandoned/` since the API is now protected and there's no lightweight workaround.

The original code by @beranka is solid, but Alza's protection is too strong for an automated OSINT tool without residential proxies.

---

beranka:
Well you would typically use this tool from a residential IP and it does work there. For me in Central Europe it worked at my parents home, my home and a hotel I stayed at... Of course, if you run it from some slopilot datacenter, then it wouldn't work, so is the case for plenty of other modules 🙃

Azaucifer:
@beranka Thanks for the clarification! That makes sense.

So the module works from residential IPs but gets blocked from datacenter IPs - that's actually expected behavior for many modules.

I already have the split files ready:
- `alza_cz.py` (Czech Republic)
- `alza_hu.py` (Hungary)
- `alza_sk.py` (Slovakia)
- `alza_at.py` (Austria)
- `alza_de.py` (Germany)

Each is self-contained as requested.
Since you're still active, I can push these to this PR as a co-author.

beranka:
@kaifcodec I did split the module into five almost identical ones as you suggested, see bc305e2614a297a170a092ca81f3cdc19c66ddbf.

Azaucifer:
@beranka Thanks for adding the split files! Since you already pushed the changes to #492, I'll close #634.

Great work!

kaifcodec:
@beranka It's likely okay.
I’m a bit busy right now, but I’ll review it soon.

kaifcodec (approved):
@beranka I reviewed the PR, but just like @Azaucifer pointed out earlier regarding the 403 blocks, I ran into the same error during my tests. 

You mentioned that it works on residential connections, but testing from a standard local home network still hits the 403 block. So the issue likely isn't the IP type (datacenter vs residential), but rather the geographic region from which the request is sent to Alza's platforms.

But as you are confirming that it really works in your region then it can be merged, however it will only be useful for the users from your region.

**Thank you for the PR, Feel free to contribute further.**

PR #494, opened 2026-08-03
fix: correct Fore.reset typo crashing update error handler

### LINKED ISSUE #493: [Bug] update_self()'s error handler crashes instead of warning the user when reinstall fails
**Summary**

`update_self()`'s own error handler crashes with an unrelated `AttributeError` instead of showing the intended "Failed to update" message, when the pip install step fails after uninstall has already succeeded.

**Where**

`user_scanner/utils/update.py`, line 22

**Details**

```python
def update_self():
    print("Updating user-scanner using pip...\n")
    try:
        subprocess.check_call([
            sys.executable, "-m", "pip", "uninstall", "user-scanner", "-y"
        ])
        subprocess.check_call([
            sys.executable, "-m", "pip", "install", "user-scanner"
        ])
    except subprocess.CalledProcessError as e:
        print(f"{Fore.RED}Failed to update user-scanner: {e}{Fore.reset}")
        return
```

`Fore.reset` is lowercase. Colorama's `Fore` class only exposes `Fore.RESET` (uppercase), `Fore.reset` doesn't exist.

**Reproduction**

Mocked `subprocess.check_call` so uninstall succeeds and install raises a real `subprocess.CalledProcessError` (matching what pip actually raises on failure), then called `update_self()`:

Updating user-scanner using pip...

Traceback (most recent call last):
...
File ".../user_scanner/utils/update.py", line 22, in update_self
print(f"{Fore.RED}Failed to update user-scanner: {e}{Fore.reset}")
^^^^^^^^^^
AttributeError: 'AnsiFore' object has no attribute 'reset'

**Impact**

`update_self()` uninstalls first, then reinstalls, with no rollback if install fails, for example if the user is offline, which is also the most likely reason a version check would have failed in the first place, see #480. The `except` block exists specifically to warn the user when this happens, but it crashes before the message is printed, so the user gets a raw traceback with no indication that the tool is now uninstalled.

**Suggested fix**

Change `Fore.reset` to `Fore.RESET` on line 22.

I would like to work on this myself.

A-S-Manoj:
Fixes #493 

**Problem**

`update_self()`'s error handler uses `Fore.reset` (lowercase), which doesn't exist on colorama's `Fore` class (only `Fore.RESET` does). When `pip install` fails after `pip uninstall` has already succeeded, the warning meant to tell the user the tool is now uninstalled crashes instead with `AttributeError: 'AnsiFore' object has no attribute 'reset'`, leaving them with a raw traceback and no indication of what happened.

**Fix**

Changed `Fore.reset` to `Fore.RESET` on the affected line.

**Changes**

- `user_scanner/utils/update.py` — one-character fix
- `tests/test_update.py` — added a regression test that mocks uninstall succeeding and install failing with a real `subprocess.CalledProcessError`, and asserts the error message prints without raising

**Testing**

Confirmed the new test fails on unmodified code with the exact real `AttributeError`, and passes with the fix. Full suite: 173 passed, 3 skipped (pre-existing, unrelated) against current `main`. `ruff check .` and `mypy user_scanner` both clean.

### DIFF
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -1,5 +1,6 @@
 from user_scanner.core import helpers as hl
-
+from user_scanner.utils import update as upd
+import subprocess
 
 def test_default_config():
     configs = hl.load_config()
@@ -29,3 +30,20 @@ def test_config_set(tmp_path, monkeypatch):
 
     hl.save_config_value("auto_update_status", True)
     assert get_status() is True
+
+def test_update_self_prints_message_on_install_failure(monkeypatch, capsys):
+    calls = []
+
+    def fake_check_call(cmd, *a, **kw):
+        calls.append(cmd)
+        if "uninstall" in cmd:
+            return 0
+        raise subprocess.CalledProcessError(1, cmd)
+
+    monkeypatch.setattr(subprocess, "check_call", fake_check_call)
+
+    upd.update_self()
+
+    out = capsys.readouterr().out
+    assert "Failed to update user-scanner" in out
+    assert len(calls) == 2  # uninstall attempted, then install attempted
--- a/user_scanner/utils/update.py
+++ b/user_scanner/utils/update.py
@@ -19,7 +19,7 @@ def update_self():
             sys.executable, "-m", "pip", "install", "user-scanner"
         ])
     except subprocess.CalledProcessError as e:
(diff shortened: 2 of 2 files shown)


kaifcodec (approved):
LGTM @A-S-Manoj.
And thanks for the test update.

A-S-Manoj:
Appreciate the quick review.

Issue #491, opened 2026-08-02
Linux connection error


VamatoHD:
**Description:**
When using the tool in Linux, most of the modules result in `Error (ConnectError: [errno -2] name or service not known)`. Since the module performs many requests at once, this may be DNS hitting the rate limit.

I'm not able to reproduce this in Windows, only in Linux (only tested in Linux Mint).

**Possible solutions:**
Perhaps lower `MAX_CONCURRENT_REQUESTS` on Linux.

kaifcodec:
@VamatoHD I mean, try testing it with a lower `MAX_CONCURRENT_REQUESTS` value using the `-C` flag and see if the same issue still occurs.

I've run into similar issues before, such as intermittent DNS errors. In my case, they were usually caused by the OS or network stack just waking up from an idle state. After a few consecutive requests, everything started working normally again.

So it would be worth checking whether lowering the concurrency makes the issue disappear.

First test it before we update the default value.

VamatoHD:
@kaifcodec reducing it to 10 (`-C 10`) did, in fact, remove the connection error and, in my opinion, make the search faster.

I believe that reducing it allowed the DNS to work properly, and every request was made on time, whereas, with 60 threads, most would wait for the DNS or even fail. 

I've only tested this on Linux Mint (I can't test on Windows currently).

kaifcodec:
Yeah, I think we may need to introduce platform-specific default configurations.

In my testing, it worked well on Windows 11, Ubuntu, and Termux overall. The only occasional issues I noticed were on Ubuntu and Termux, similar to what I mentioned earlier, intermittent DNS/network-related errors that usually resolved themselves after a few consecutive requests.

Anyways, do you have any ideas for addressing this? For example, are you thinking about introducing platform-specific default configurations (just for Linux Mint) or something else?

By the way did you test it that whether it get fixed after consecutive runs? Just like my case i.e. First run of the tool hit the error and the second or maybe 3rd run goes perfect?

VamatoHD:
In my testing, it is also intermittent, but since there are so many workers, it ends up erroring many modules. And it's basically the same (just different modules) on each run.

Implementing platform-specific wouldn't work, as it can be hardware-specific.

We could keep a list of modules that failed due to a connection error (the DNS error as mentioned early) and try them again, maybe just by using a queue.

kaifcodec:
[...]

PR #496, opened 2026-08-03
Refactor: separate media from extra

### LINKED ISSUE #485: Refactor all scanner modules to use the explicit `media` dictionary for images
## Description
In PR #471, we introduced a dedicated `media` dictionary inside the `Result` dataclass to handle extracted image URLs (like avatars, banners, and snapcodes) independently of the standard `extra` metadata dictionary. 
Currently, many legacy scanner modules still extract image URLs directly into the `extra` dictionary (e.g., `extra["avatar"] = "https..."`). While the PDF generator has a fallback heuristic to catch these, this approach is fragile and clutters the raw metadata text tables.
We need to migrate all remaining modules to populate image URLs inside a `media` dictionary instead of `extra`.
## Acceptance Criteria
- [ ] Audit all modules in `user_scanner/user_scan/` and `user_scanner/email_scan/` for image extractions (search for `"avatar"`, `"image"`, `"pfp"`, `"profile_picture"`, etc.).
- [ ] For every match, initialize a `media = {}` dictionary alongside `extra = {}`.
- [ ] Move the image URL assignments from `extra` to `media` (e.g., change `extra["avatar"] = url` to `media["avatar"] = url`).
- [ ] Pass the new `media` dictionary to the `Result` object (e.g., `Result.taken(extra=extra, media=media)`).
## Examples
See the recent commit [here] migrating `github.py` and `steam.py` for a clear example of the required changes.
```python
# BEFORE
extra = {}
if avatar_url: 
    extra["avatar"] = avatar_url
return Result.taken(extra=extra)
# AFTER
extra = {}
media = {}
if avatar_url: 
    media["avatar"] = avatar_url
return Result.taken(extra=extra, media=media)

VamatoHD:
This PR implements the changes stated in #485 ( closes #485 )

Hopefully every module that needed the changes is covered, but it's better to check again.
I ran every module with various usernames/emails, and I didn't find any module that still uses extras for media.


kaifcodec:
@VamatoHD I will review it slowly, if you can then take the issue #491 in this meantime.

kaifcodec:
~@VamatoHD The tests are failing, I saw that you fixed the missing field in `as_dict()` function but still why is it failing?~
~Can you check the tests?~

kaifcodec:
@VamatoHD I just checked, and the issue is on line 22 of `result.py` in `CSV_FIELDS`.
And also ruff is complaining about unused import.

VamatoHD:
@kaifcodec Working on it, not sure how I let it pass...

kaifcodec:
No worries then, leave that part. I will fix that when I am going to review.

kaifcodec:
@VamatoHD Check this out. I refactored some of the missing ones like in `email_scan/social/gravatar.py`and in gitlab the `gravatar_url` represents a image so refactored it too and some others.

And I think all are covered now.

VamatoHD:
@kaifcodec I'm checking, and CSV and JSON display the result's media; however, PDF doesn't appear to, only the profile pictures. Is this intentional?

kaifcodec:
Oh, that's not intentional, the pdf export feature merged earlier so it wasn't fixed, so it will be better that when generating pdf report we add the media dictionary to `extra` so that image URLs doesn't get missed to show up in `EXTRACTED DEEP INTELLIGENCE` at the end of the pdf report.

kaifcodec (approved):
LGTM.

json-hunter07 (approved):

PR #497, opened 2026-08-03
fix: motherless.xxx dns resolution hanging

kristoisberg:
## Summary

  Fix Motherless scans hanging for roughly two minutes when DNS resolution fails.

  ## Root cause

  The existing 5-second HTTPX timeout did not cover synchronous DNS resolution. Python’s blocking getaddrinfo() could therefore continue retrying long after the configured request timeout before eventually returning a DNS error.

  ## Fix

  - Route Motherless requests through curl_cffi, whose request timeout covers the DNS lookup path.
  - Apply the configured global timeout, falling back to 5 seconds.
  - Enable curl’s QUICK_EXIT option on shared impersonated sessions to avoid waiting on resolver-thread cleanup during shutdown.
  - Update the session test double to accept the new curl option.

  This keeps the existing error handling and profile lookup behavior while ensuring broken DNS cannot stall the scan.

### DIFF
--- a/tests/test_impersonate.py
+++ b/tests/test_impersonate.py
@@ -7,9 +7,10 @@ from user_scanner.core.result import Result, Status
 class FakeSession:
     instances = []
 
-    def __init__(self, impersonate=None, proxies=None):
+    def __init__(self, impersonate=None, proxies=None, curl_options=None):
         self.impersonate = impersonate
         self.proxies = proxies
+        self.curl_options = curl_options
         self.calls = []
         FakeSession.instances.append(self)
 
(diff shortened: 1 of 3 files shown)


kristoisberg:
The issue was caused by a combination of DNS behavior and local resolver configuration, rather than Motherless request retries.

motherless.xxx currently returns NXDOMAIN. On the affected machine, systemd-resolved was configured with mandatory DNS-over-TLS while also accepting ISP DNS servers that did not support DNS-over-TLS. Cloudflare returned NXDOMAIN immediately, but systemd-resolved waited for the unreachable ISP resolvers to finish retrying before returning the negative result. That took about 120 seconds. Python’s synchronous getaddrinfo() call blocked for that entire period, outside HTTPX’s five-second timeout.

It appeared specific to Motherless because other scanned domains returned positive DNS answers, which systemd-resolved could accept immediately. Any other nonexistent domain could have triggered the same delay. Users without the conflicting DNS-over-TLS configuration received the NXDOMAIN response immediately, explaining why the problem was not reproducible everywhere.

PR #499, opened 2026-08-03
fix: LinkTree username module errors

kristoisberg:
This PR fixes two issues regarding in the LinkTree username module:
- Uses impersonate_validate to bypass TLS fingerprinting, which started causing 403 responses.
- Expands 404 page validations, because apparently LinkTree (now?) has two different 404 pages depending on whether the username does not conform to their requirements, or does merely not exist, and the checks previously only checked the former.

Examples of the different 404 pages:
https://linktr.ee/invalid-username
https://linktr.ee/validusername1

### DIFF
--- a/user_scanner/user_scan/creator/linktree.py
+++ b/user_scanner/user_scan/creator/linktree.py
@@ -34,6 +34,12 @@ def validate_linktree(user: str) -> Result:
         if response.status_code != 200:
             return Result.error(f"Unexpected response status: {response.status_code}")
 
+        banned_user = data.get("query", {}).get("username")
+        if data.get("page") == "/status/blocked" and (
+            isinstance(banned_user, str) and banned_user.lower() == user.lower()
+        ):
+            return Result.taken(extra={"banned": True})
+
         account = page_props.get("account", {})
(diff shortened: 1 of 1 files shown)


kaifcodec:
@kristoisberg I tested this changes but I only got,

`  [!] Linktree [https://linktr.ee/mia] (mia): Error (Timeout: Failed to perform, curl: (28) connection timed out after 15002 milliseconds. see https://curl.se/libcurl/c/libcurl-errors.html first for more details.)`  -> this one using CA VPN.

or 
`  [!] Linktree [https://linktr.ee/mia] (mia): Error (Linktree profile markers were missing)` -> this one when no VPN used.

Errors.

However old one with httpx is working fine only issue with that  is it's unable to extract the `extra` data.

kristoisberg:
@kaifcodec Could you retest? The account with the name "mia" is banned, and banned users were not handled by the module. I'm unable to comment on the VPN timeout, though.

Interesting that you didn't run into the 403 errors without impersonation. Perhaps they only occur regionallly? Similarly, the NexusMods module was working for me when I implemented it, but required impersonation on the reviewer's side.

kaifcodec (approved):

PR #500, opened 2026-08-03
fix: follow redirects in Zomato username module

kristoisberg:
This PR enables redirects in the Zomato username module. At least in my case, the site redirects to a localised version (i.e. https://www.zomato.com/pl/.../reviews), causing the module to currently fail due to the 302 response code.


kaifcodec (approved):
@kristoisberg This one also works fine for me I am not getting 302 but sometimes when region changes, it needs redirect, so it's always safe to redirect.
so merging it.

PR #504, opened 2026-08-04
fix: Steam user module errors

kristoisberg:
Another module that requires impersonation on my end to work properly... It previously returned with the correct "not found" error page when the user was not found, but with 429 in case the user existed.

### DIFF
--- a/user_scanner/user_scan/gaming/steam.py
+++ b/user_scanner/user_scan/gaming/steam.py
@@ -1,7 +1,9 @@
+import html
+import re as local_re
+
 from user_scanner.core.impersonate import impersonate_validate
 from user_scanner.core.result import Result
-import re as local_re
-import html
+
 
 def validate_steam(user):
     url = f"https://steamcommunity.com/id/{user}/"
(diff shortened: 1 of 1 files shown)


brunolm (approved):
Tested this live on both `main` and this branch — same results either way:

```
brunolm              -> Found  (steam_id 76561198036291369, persona_name, real_name, summary)
zzznotarealuser99xzq -> Not Found
```

So I can't reproduce the 429 — httpx isn't rate-limited from where I am, which suggests it's IP/region dependent.

One unrelated thing while you're in the file: the `200` branch returns `Result.taken()` from a bare `else`, which goes against `CONTRIBUTING.md` rule #1 for `user_scan` modules — *"You MUST explicitly verify a unique string or JSON key for BOTH the `taken` and `available` states. Never use a bare `else: return Result.available()` block."* You already extract `steamid` — gating on it instead gives a positive marker for "found":

```python
if "Error</title>" in response.text:
    return Result.available()

steamid_match = local_re.search(r'"steamid"\s*:\s*"((?:[^"\]|\.)*)"', response.text)
if not steamid_match:
    return Result.error("Unrecognized response")

extra = {"steam_id": steamid_match.group(1)}
```

with the rest of the extraction following, and no `else` wrapper.

kristoisberg:
@brunolm Thanks for the suggestion, I will make the change.

The fingerprint checks seem to often be region-based indeed: for example, NexusMods was the opposite - httpx worked for me, but didn't for you.

kaifcodec (approved):
It's perfect now, LGTM.

PR #505, opened 2026-08-04
chore: add CLAUDE.md/AGENTS.md contributor guide and organize .gitignore

brunolm:
## What

- Add **CLAUDE.md** and **AGENTS.md** (kept in sync) — a short guide orienting AI assistants and new contributors. It points to `README.md`, `CONTRIBUTING.md`, and the `docs/` files, then summarizes the module conventions (validator signature, `Result` returns, no-false-positive rules, transport choice, namespace mapping) and the local CI gates (`ruff`, `mypy`, `pytest`).
- Organize **.gitignore** into logical groups and add ignores for generated/private files: `output/`, `uv.lock`, `docs/local/`, `*.local.*`, and `.playwright-mcp/`.

## Why

Lowers the ramp for both humans and AI tools contributing new scan modules by putting the essential conventions and doc pointers in one discoverable place, and keeps generated/local artifacts out of version control.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kristoisberg:
Great idea, I've been thinking of doing this too. Some thoughts:
- Rather than duplicating the contents, CLAUDE.md could reference AGENTS.md by simply setting the contents as `@AGENTS.md`.
- I would add an explicit instruction not to add unit tests to individual modules.

Somewhat irrelevant here, but could be useful: I have had great success with the following tools when working on user-scanner:
- The https://github.com/dietrichgebert/ponytail skill - fits well with our philosophy of keeping modules simple and compact
- https://ax.yusuke.run/ - very efficient for investigating page structures and behaviour

brunolm:
Thanks, great suggestions, all added.

On the same topic, two tools of my own that I use for this kind of investigation:

- [stealth-fetch](https://github.com/brunolm/stealth-fetch-mcp): MCP server that fetches with a real browser TLS fingerprint (curl_cffi, same approach as our impersonating transport)
- [patchright-mcp](https://github.com/brunolm/patchright-mcp): fork of playwright-mcp using patchright, for browsing bot-walled sites

I didn't include these in the docs because they're my personal tools and I don't want to give any guarantees that they'll keep working forever, but if anyone is interested feel free to use them.

kristoisberg (approved):

kristoisberg (changes requested):
I would suggest adding a directive for the agent to learn how to use ax (if it's available), similarly to their website:
> Run `ax agent-context` to learn it

Other than that, great additions!

kristoisberg (approved):
lgtm

PR #509, opened 2026-08-05
fix: Fansly user module errors

kristoisberg:
This PR fixes the Fansly username module erroring out in case of an available username. The module previously checked for a 404 response code, but the API actually responds with 200, with the response array being empty.


json-hunter07 (approved):
It's perfect LGTM.
I tested it with a known username and it worked perfect!

kaifcodec:
Thanks for the review and testing @json-hunter07 
**MERGING IT!**
Thanks for the fix @kristoisberg

PR #510, opened 2026-08-05
fix: Blogger username module errors

kristoisberg:
Another module that required following redirects to work properly, at least in my case.


kaifcodec (approved):
Yeah, I also noticed it just today that it was returning 302 for a positive hit.
Merging it!

PR #511, opened 2026-08-05
feat: auto enable show_all when scanning <= 10 modules with -m flag

kaifcodec:
- Automatically enable show_all when 10 or fewer modules are requested via the -m flag
- Ensure results for targeted module scans are displayed on the terminal instead of exiting silently
- Retain default found-only behavior when scanning more than 10 modules unless --all is specified

### DIFF
--- a/user_scanner/__main__.py
+++ b/user_scanner/__main__.py
@@ -354,10 +354,16 @@ def main():
                 )
 
     results = []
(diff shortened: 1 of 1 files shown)


json-hunter07 (approved):
@kaifcodec This is really a cool feature!
LGTM!

PR #514, opened 2026-08-05
feat(linkedin): enrich username profiles from the link-preview JSON-LD

brunolm:
## What

Turns the `linkedin` username module from existence-only into a rich enrichment module. LinkedIn's Twitterbot link-preview page — already used here to bypass the anti-bot `999` response — embeds a JSON-LD `Person` node plus OpenGraph and profile meta tags. Parsing those, a hit now returns:

`name`, `pronouns`, `headline`, `bio`, `badges`, `location`, `country`, `experience`, `education`, `languages`, `awards`, `articles` (with publish dates), `last_posted`, `similar_profiles` (name + scannable handle), `followers`, `connections`, and the `avatar` in `media`.

All of it comes from the single request the module already makes — no extra fetches, no credentials, same transport.

## How

- Following redirects lets country-hosted vanity handles (e.g. `br.linkedin.com`) enrich too; missing handles still `404` without redirecting, so the found/not-found signal is preserved.
- Found is confirmed by the `og:type=profile` marker and not-found by the explicit `Profile Not Found` title — never a bare status code. Anything else returns `Result.error` (e.g. the `999` rate-limit stays a non-verdict).
- Images go in `media`; everything else in `extra`, following `CONTRIBUTING.md`.

## Edge cases handled (found live-testing real handles)

- HTML-encoded bios (`&gt;`, `<br>`) are decoded.
- Sparse profiles with no JSON-LD `Person` node: `name` falls back to the `profile:first_name`/`profile:last_name` meta tags.
- The headline-less `<name> - <location> | Professional Profile` boilerplate is suppressed instead of being reported as a headline.
- Duplicate awards are de-duplicated.
- Pronouns appended to the name (`(She/Her)`) are split into their own field, gated on the slash so real parentheticals like `(PhD)` are left intact.
- `uk` host normalized to ISO `GB` for a consistent `country` code.

## Testing

- `ruff check .` and `mypy user_scanner` clean.
- `pytest` passes (the two `*_file_unreadable` failures are pre-existing and Windows-only — they `chmod` a file unreadable, a no-op on this platform).
- Live-tested against a real handle and a nonexistent one, plus ~20 real profiles across BR/US/GB/CA/DE/SE/NL/IN/CH/FI to verify country handling, sparse profiles, accented handles, and every field.

### DIFF
--- a/user_scanner/user_scan/social/linkedin.py
+++ b/user_scanner/user_scan/social/linkedin.py
@@ -1,19 +1,214 @@
-from user_scanner.core.orchestrator import status_validate
+import html
+import json
+import re
+
+import httpx
+
+from user_scanner.core.orchestrator import generic_validate
 from user_scanner.core.result import Result
+
+# LinkedIn whitelists social-media crawler bots for link previews, bypassing the
+# anti-bot 999 response that blocks regular user agents. The preview HTML carries
+# a JSON-LD Person node plus OpenGraph tags, which is all the profile metadata we
+# surface. Note: bulk scanning may still trigger rate limiting (999) after a few
+# rapid requests — that returns Result.error, never a verdict.
+HEADERS = {"User-Agent": "Twitterbot/1.0"}
 
 
 def validate_linkedin(user: str) -> Result:
-    # LinkedIn whitelists social media crawler bots for link previews,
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@brunolm LGTM.
Nice work!

PR #515, opened 2026-08-06
fix(tiktok): detect accounts via the embed page instead of a dead marker

brunolm:
### tl;dr

- **The module reported every handle as Found**, including ones I invented — a bare `else` behind a body marker that no longer exists.
- **It was also nondeterministic** — the same nonexistent handle returned `TAKEN` and then `AVAILABLE` from unmodified code.
- **Fixed by switching to `tiktok.com/embed/@<user>`**, which server-renders profile data for public accounts.
- **Adds 7 metadata fields plus an avatar** where the module previously returned nothing.
- **A 400 now returns `error`, not `available`** — a private account is indistinguishable from an unregistered handle, verified on a live account toggled private.
- **`heartCount` overflows int32** and needs unwrapping — the subtlest thing in this diff.
- **`videoList` deliberately excluded** — it's a recent-videos slice, not a total.
- **Gates clean**, tested live across public, private, nonexistent and invalid handles.

---

### The module reported every handle as Found

Existence was decided by scraping the profile page:

```python
if 'statuscode":10221' in response.text.lower():
    return Result.available()
else:
    return Result.taken()
```

That page is now a 1462-byte client-rendered shell, byte-identical for real and missing handles, with no `__UNIVERSAL_DATA_FOR_REHYDRATION__` and no `SIGI_STATE`. The marker never matches, so the `else` always wins:

```
brunolm_7            -> TAKEN
zzznotarealuser99xzq -> TAKEN   # invented
xqzznobodyhere4471   -> TAKEN   # invented
```

This is the bare-`else` pattern `AGENTS.md` rule 3 prohibits. It rotted silently because nothing fails loudly when a body marker disappears.

### It was also nondeterministic

The marker appears intermittently rather than never. Across runs, `zzznotarealuser99xzq` returned `TAKEN` once and `AVAILABLE` later from unmodified code. That is worse than a consistent failure: re-running can appear to corroborate whichever answer you got first.

### Fixed by switching to the embed page

`tiktok.com/embed/@<user>` server-renders a `userInfo` blob for public accounts. A plain httpx request through `generic_validate` is enough — no impersonation needed.

`userInfo` is extracted with a brace-balanced scan rather than a regex terminating at `,"videoList"`, so accounts with no videos still parse.

### Adds 7 metadata fields plus an avatar

`username`, `name`, `user_id`, `bio`, `followers`, `following`, `likes`, plus `verified` / `private` when true, and the avatar in `media`. `extra` was previously empty on every hit.

### A 400 now returns `error`, not `available`

TikTok answers 400 for a private account exactly as it does for an unregistered handle. I confirmed this against a live account toggled private mid-test, and no public endpoint separates the two:

| Endpoint | Private | Nonexistent | Discriminates? |
| --- | --- | --- | --- |
| `/embed/@user` | 400 | 400 | No |
| `/oembed?url=` | 400 | 400 | No |
| `/@user` (curl_cffi) | 200, 1462 B | 200, 1462 B | No — byte-identical |
| `/api/user/detail/` | 200, empty | 200, empty | No |
| `/node/share/user/@` | 403 | 403 | No |
| `/@user` as Googlebot | 403 | 403 | No |
| signup-style check endpoints | `"url doesn't match"` | same | No |

Returning `available` would assert a handle is free when it may belong to a private account — an estimated 20–25% of TikTok accounts. This follows `erome.py`, which refuses a verdict on a 410 for the same reason.

**This trades away availability detection**: the module can no longer return `available` from any network path. That is deliberate, and it is the cost of not asserting something false.

### `heartCount` overflows int32 — review this closely

Large accounts return a **negative** like count:

```
khaby.lame -> heartCount: -1642920348
-1642920348 + 2**32 = 2652046948   # matches his real ~2.65B
```

`_unwrap_int32` corrects it, applied to all three counters since followers can cross the same threshold. Small accounts are unaffected, so this is invisible unless a large account is tested.

### `videoList` deliberately excluded

The page also carries a `videoList`, but it is the embed's recent-videos slice, not a total. Any count derived from it would be a figure whose label I cannot trace.

### Gates clean, tested live

`ruff check .` clean · `mypy user_scanner` clean across 435 files · `pytest` 172 passed (2 failures are pre-existing Windows chmod no-ops, unrelated).

Verified against a public account, the **same account toggled private**, a large verified account (`khaby.lame`), two nonexistent handles, and every invalid-syntax branch. Verdicts were stable across repeated runs.

Behaviour across the privacy toggle, same handle and command:

```
private -> [!] Tiktok (brunolm_7): Error (No public account (a private one looks the same))
public  -> [✔] Tiktok (brunolm_7): Found  (username, name, user_id, bio, followers, following, likes)
```

**Note on visibility:** the default terminal view prints neither errors nor available results, so a private account still shows `No results found` unless `--all` is passed. The corrected status and reason always reach JSON/CSV export. Surfacing errors by default would mean changing the shared formatter, which is out of scope here.

### DIFF
--- a/user_scanner/user_scan/social/tiktok.py
+++ b/user_scanner/user_scan/social/tiktok.py
@@ -1,3 +1,4 @@
+import json
 import re
 
 from user_scanner.core.helpers import get_random_user_agent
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@brunolm Thanks!
**MERGING IT!**

kristoisberg:
@brunolm Just curious, when did this break?

brunolm:
I don't know, I found out when I was trying to improve the scan.

kristoisberg:
The strange thing is that when I revert this commit, the module detects private accounts correctly, and distinguishes them from non-existent accounts...

kaifcodec:
@kristoisberg Maybe regional difference in response? Did you inspect the response body for that?

kristoisberg:
Indeed, the source code is entirely different when I access TiKTok through a VPN near where @brunolm is located. Even the tab name couldn't access user data!

image

I have three thoughts:
a) Perhaps this is an ongoing change given the regional differences and that even the tab name is incorrect. Maybe they will expose something we could use in the long run.
b) We could use a hybrid approach, falling back to the old one, and only reporting "available" if neither approach couldn't find a user.
c) Time for Playwright, after all?

Edit: From what I can see, `/api/user/detail/` is not empty for private accounts when accessed from an actual browser through the frontend.

kristoisberg:
I tried a few different approaches and the result of this PR is the best we've currently got.

- `/api/user/detail/` only returns non-empty results with a valid verification signature. A few efforts have been made to reverse engineer and produce these signatures (the most relevant being https://github.com/alana72212/TikTok-Signer), but TikTok keeps updating the signature algorithms so they quickly become invalid. I tried reverse engineering the current signatures with Codex for several hours, but ultimately gave up as the module would have become unmaintainable in the long run.
- The old SSR page is intermittently available when impersonating older versions of Chrome, but it requires multiple retries for the results to be consistent enough. I assume it's being phased out so it's not viable as a long-term solution.

PR #516, opened 2026-08-06
feat(gravatar): extract links, interests and profile details

brunolm:
## tl;dr

- **`job_title` was silently dropped for every profile that sets one.** The extractor read `jobTitle`; the legacy endpoint this module calls returns `job_title`.
- **`verified_accounts` was incomplete and nothing revealed it.** Both JSON APIs omit providers their schema predates — a YouTube connection is absent from each yet rendered on the page.
- **Merging the two sources double-listed accounts and reported a help page as an account URL.** Both found by scanning 11 live profiles, both fixed.
- **Page data is selected without CSS classes**, so a restyle does not silently truncate the result.
- **Links, interests and advanced details exist only in the rendered page**, so the module now scrapes it — one extra request per hit.
- **`phoneNumbers`, `contactInfo`, `pronouns`, `pronunciation` and the profile background were already in the payload and simply unread.**
- **Not verified: `crypto_addresses`, `websites`, and the MD5 fallback leg.**

Net effect on a live scan: 8 `extra` fields before, 19 after.

## `job_title` was silently dropped for every profile that sets one

```python
if entry.get("jobTitle"):          # legacy endpoint returns job_title
    extra["job_title"] = ...
```

The v3 API spells it `jobTitle`, the legacy endpoint spells it `job_title`. The key was never present, so the field vanished for every profile. Now reads either spelling.

## `verified_accounts` was incomplete and nothing revealed it

The advanced-details panel reports a count, and on a test profile it disagreed with the list:

```
verified connections (page count):  5
verified_accounts (legacy JSON):    X, LinkedIn, GitHub, Stack Overflow   -> 4
verified_accounts (v3 API):         X, LinkedIn, GitHub, Stack Overflow   -> 4
verified_accounts (page card):      + YouTube                             -> 5
```

Neither JSON source is complete, so the page card is merged onto the JSON list.

The count itself is no longer emitted — it was the length of the list beside it. It was also the only signal that the list was short, so nothing detects a future gap automatically now.

## Merging two sources double-listed accounts and leaked a help URL

Scanning 11 live profiles instead of one turned up two faults the primary test account could not expose:

| Profile | Symptom | Cause |
| --- | --- | --- |
| `ryan@boren.me` | 11 accounts listed, page says 7 | JSON stores vanity URLs (`ryanboren.link/x`), the page stores canonical ones (`x.com/rboren`); URL comparison cannot match them |
| `skeltoac@gmail.com` | `Facebook: https://support.gravatar.com/…#facebook` | Gravatar stores its own help page as the URL for providers that forbid public profile links |

Dedup now compares the service name as well as the URL, and help-page URLs are dropped — such an account is reported as `Facebook (verified)` with no URL, since the fact of the connection is real but the URL says nothing about the owner.

Every scanned profile now agrees with the page's own count:

```
brunolm7 5=5   beau 4=4   matt 0=0   nacin 2=2   skeltoac 4=4   jack 0=0
ian      7=7   jason 4=4  ryan 7=7   conrad 0=0  zach 0=0
```

## Page data is selected without CSS classes

| Region | Anchored on |
| --- | --- |
| verified accounts | the checkmark's link to `support.gravatar.com/profiles/verified-accounts`, then the account anchor's `rel="me"` |
| links | `rel="me"` anchors whose `title` differs from their `href` — a personal link carries the owner's caption, every other `rel="me"` anchor repeats its URL |
| interests | the `<h2>Interests</h2>` heading and the `<ul>` after it |
| advanced details | `<span>Label: value</span>` — no classes to begin with |

`rel="me"` is the microformats identity relation and the help URL is user-facing documentation; neither is a styling hook. Verified by rewriting every `g-profile__`, `card-item__` and `g-profile-card__` class in the live HTML and re-extracting: identical output on all 11 profiles.

Two fragilities remain rather than being removed. A link captioned with its own URL is indistinguishable from a verified account and is dropped, and `interests` now depends on the English heading instead of a class — the same kind of dependency advanced-details already had.

## Where each field comes from

| Source | Fields |
| --- | --- |
| Legacy JSON | username, display_name, profile_url, bio, location, job_title, company, pronouns, pronunciation, contact_info, phone_numbers, public_emails, background_color, `media`: avatar, thumbnail, photos, background |
| Profile page | verified_accounts (merge), links, interests, profile_type, created, last_updated, languages, timezone |

An ablation over three live profiles confirms the split is minimal: legacy JSON alone reaches 14 of 19 fields, the page alone reaches 8, together they reach all of them.

Gravatar also serves a markdown export that respells `En, Pt` as `English (primary), Portuguese (secondary)` and `GMT-03:00` as `America/Sao_Paulo`. It is deliberately not fetched — it contributes no field the page lacks, and a third request per hit is not worth two prettier values. The consequence is that `timezone` is a UTC offset, correct only outside DST.

## The scraping will rot, and it degrades quietly

Enrichment goes from 2 GETs to 3, and both enrichment fetches are wrapped in try/except and can only add to `extra`, so the Registered/Not-Registered verdict — still resting solely on the avatar endpoint's 200/404 — is untouched by this PR.

The failure mode is not symmetric: if the page markup changes in a way the new anchors do not survive, `verified_accounts` falls back to the incomplete JSON list rather than going empty, which looks identical to a profile that genuinely has fewer connections.

## Also in this PR

- **The global `-t` flag is now honoured.** The module hardcoded `timeout=15.0`. With `-t 0.001`: before `Registered`, after `Error: Connection timed out`.
- **Background image moved to `media`** alongside the avatar, following #496; `background_color` stays in `extra`.
- **Omitted deliberately:** `avatar_alt_text`, `hide_default_header_image` and `section_visibility` are rendering hints for Gravatar's own page, not facts about the account. Wallet/payments and profile feeds need an API key.

## Testing

Eleven live profiles plus four misses:

```
brunolm7@gmail.com               Found      3 GETs  extra=19 media=3  verified=5
beau@automattic.com              Found      3 GETs  extra=18 media=4  verified=4
ian@mckellar.org                 Found      3 GETs  extra= 9 media=4  verified=7
ryan@boren.me                    Found      3 GETs  extra=15 media=3  verified=7
skeltoac@gmail.com               Found      3 GETs  extra=11 media=3  verified=4
jasonhoffman@gmail.com           Found      3 GETs  extra=10 media=3  verified=4
nacin@wordpress.org              Found      3 GETs  extra=10 media=3  verified=2
conrad@trueventures.com          Found      3 GETs  extra=10 media=3  verified=0
jack@baty.net                    Found      3 GETs  extra= 9 media=3  verified=0
zach@londonmade.co.uk            Found      3 GETs  extra= 6 media=3  verified=0
matt@mullenweg.com               Found      3 GETs  extra= 5 media=3  verified=0
zzznotarealuser99xzq@gmail.com   Not Found  2 GETs
john@example.com                 Not Found  2 GETs
not-an-email                     Not Found  2 GETs
""                               Not Found  2 GETs
```

The set covers a background image and colour, phone numbers, a profile with no verified accounts and no links, a near-empty profile, and both faults described above. `  BrunoLM7@GMAIL.com ` (case and surrounding whitespace) returns the same 19 fields as the clean spelling.

Not verified: `crypto_addresses` and `websites` (no public profile found that sets them), and the MD5-hash fallback leg, unchanged in behaviour but refactored into a loop alongside the SHA-256 attempt.

### DIFF
--- a/user_scanner/email_scan/social/gravatar.py
+++ b/user_scanner/email_scan/social/gravatar.py
@@ -93,18 +93,23 @@ async def _collect_profile(client: httpx.AsyncClient, email_hash: str) -> tuple[
                 _extract_profile_data(entries[0], extra, media)
     except Exception:
         pass
+    page_read = False
     try:
         response = await client.get(f"{PROFILE_HOST}/{email_hash}", headers=HEADERS)
         if response.status_code == 200:
             _extract_page_data(response.text, extra)
+            page_read = True
     except Exception:
         pass
-    try:
-        response = await client.get(f"{PROFILE_HOST}/{email_hash}.md", headers=HEADERS)
-        if response.status_code == 200:
-            _extract_markdown_data(response.text, extra)
-    except Exception:
-        pass
+    # The export only refines the two fields the page already reports, so it is
+    # worth a request only once one of them is known to be set.
+    if not page_read or any(key in extra for key in MARKDOWN_KEYS.values()):
+        try:
+            response = await client.get(f"{PROFILE_HOST}/{email_hash}.md", headers=HEADERS)
+            if response.status_code == 200:
+                _extract_markdown_data(response.text, extra)
+        except Exception:
+            pass
     return extra, media
 
 
--- a/user_scanner/email_scan/social/gravatar.py
+++ b/user_scanner/email_scan/social/gravatar.py
@@ -1,6 +1,7 @@
 import hashlib
 import html
 import re
+from collections.abc import Iterator
 
 import httpx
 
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@brunolm Tested it myself and it's impressive now, with the amount of data it extracts.
**MERGING IT!**

PR #519, opened 2026-08-06
fix: BentBox username module errors

kristoisberg:
This PR fixes several issues with the BentBox username module:
- The response structure for non-found usernames had changed, resulting in errors.
- The content checks for profile pages were not specific enough, with "bentbox.co" being present on the page being enough to consider the username taken.


kaifcodec (approved):
LGTM.

PR #520, opened 2026-08-06
fix: resolve BDSM Singles browser challenge

kristoisberg:
BDSM Singles has introduced a browser challenge, which caused the username module to fail. This PR solves the challenge to make the module work again.

PS: Could be another regional thing. If the module previously worked for any of you, and this PR broke something, please let me know.


brunolm:
There's a conflict, but LGTM

kaifcodec:
@brunolm I think there's no need to merge it or is there? 
You already had opened a pr that has been merged #556 , Is here something in this that is needed to be merged?

brunolm:
@kristoisberg can you verify if main works for you and close this if not needed?

kristoisberg:
I resolved the merge conflicts.

@brunolm I re-added the check for a 302 error code, interpreting it as an available username. The comment in your patch described two different scenarios where a 302 error code could occur (available username and browser challenge), this PR resolves the latter, only keeping available username as a (known) option. Thoughts?

brunolm (approved):

kaifcodec (approved):

PR #521, opened 2026-08-06
fix(user_scan): restore 11 modules that could never return a verdict

brunolm:
## tl;dr

- **Three modules returned a *wrong* verdict, not an error.** `academia` said `taken` on a bare 200, `osta` called real accounts free because its lookup is case-sensitive, and any HTML-based `reddit` check reports suspended accounts as free.
- **11 of 228 username modules could never return a verdict at all.** A fingerprint bot wall 403'd every request, so every handle came back `Error`. All 11 now work.
- **`reddit` now depends on solving an undocumented proof-of-work challenge.** Every JSON endpoint hard-403s until the session is cleared.
- **`sourceforge` and `themeforest` are pinned to `impersonate="safari184"`.** Every other fingerprint, including the default, gets a Cloudflare interstitial.
- **`producthunt.py` carries the same latent false positive as `academia`, left unfixed.** The site is unreachable from here, so a replacement marker cannot be verified.
- **4 other 403'd modules are deliberately untouched** — they are IP/geo blocked, not bot-walled.
- **Zero false positives across 33 invented-handle checks**; 32 real accounts resolve with metadata.

## Three modules returned a wrong verdict, not an error

| Module | Fault | Effect |
| --- | --- | --- |
| academia | Existence read off a bare HTTP 200 | Every handle Found once unblocked |
| osta | Seller lookup is case-sensitive | Real accounts reported free |
| reddit | Suspended accounts render the free-handle page | Suspended handles reported free |

```python
if response.status_code == 200:
    return Result.taken(url=url)          # academia.py, before
```

```
kristjan123 -> Not Found                       # osta, before
kristjan123 -> Found (username: Kristjan123)   # osta, after
```

Reddit renders *suspended* accounts with the same "nobody on Reddit goes by that name" page as a free handle, so `awkwardtheturtle` is indistinguishable from an invented handle on HTML alone. The availability endpoint does not have that flaw, which is why it drives the verdict. Suspended and deleted accounts report `Found` with a `status` field, because Reddit never releases the name.

## 11 of 228 username modules could never return a verdict at all

Routing through the existing `impersonate_*` helpers clears most of them. Tumblr needed a different host: `<user>.tumblr.com` serves its own "Checking your browser..." interstitial that no HTTP client clears, while the `www` route serves the blog payload directly.

| Module | Verdict source |
| --- | --- |
| academia | embedded `ld+json` ProfilePage, canonical handle matched |
| adultism | `schema.org/Person` profile container |
| apclips | 200 + `data-username`, or redirect to `/models` as the miss |
| kick | `api/v2/channels/<user>` |
| npmjs | `/~<user>` with `x-spiferack: 1` |
| osta | `listing.seller`, with a case-resolving search pre-step |
| reddit | `/api/username_available.json` |
| sourceforge | `/rest/u/<user>/profile` (Allura REST) |
| themeforest | canonical link on the profile page |
| tumblr | `www.tumblr.com/<user>` embedded blog object |
| unsplash | `/napi/users/<user>` |

Each confirms *found* with a body marker tied to the requested handle and *not found* with an explicit marker, never a bare status code. Anything else returns `Result.error`.

## `reddit` now depends on solving an undocumented proof-of-work challenge

Reddit hard-403s every JSON endpoint from an uncleared session regardless of fingerprint or headers. Clearance is only offered through an HTML interstitial carrying a proof-of-work whose solution is its seed repeated twice, and `reddit.py` solves it to get the cookie.

- **Fails safe** — an unrecognised challenge shape returns `Result.error`, never a verdict.
- **Cached per process behind a lock**, so a multi-handle scan pays once: over 6 handles that is 3 homepage fetches (~700KB) instead of 6 (~2.3MB).
- **Empty-body 404s are retried, populated ones are not.** Reddit sheds requests under burst with an empty 404 rather than a 429, which the old code read as "no verdict". `_name_available` retries only on an empty body (3 attempts, 0.5s incremental backoff); a populated body is Reddit genuinely answering (e.g. `BAD_USERNAME`) and returns immediately, so invalid handles still yield `Error` instead of being retried into a verdict.

## `sourceforge` and `themeforest` are pinned to `impersonate="safari184"`

Both are 403'd under chrome, chrome131, chrome136, firefox133, firefox135, edge101 and safari180. Forcing them back to `chrome` confirms a challenge never becomes a verdict:

```
SF brondsem             -> Error | Blocked by a Cloudflare challenge
TF zzqnotarealuser8813x -> Error | Blocked by a Cloudflare challenge
```

SourceForge's `/u/` path also rate limits with that same challenge body under a 429, so the check keys on the title, not the status. The verdict lives on `/rest/u/`; the `/u/` fetch only enriches and can never change it.

## `producthunt.py` carries the same latent false positive, left unfixed

It has the pattern `academia` had — `taken` on a bare 200 — so it will report every handle as Found if it ever clears Cloudflare. I cannot reach the site to verify a replacement marker, and guessing one is how a false positive gets written. Flagged here so it is known rather than a surprise.

## 4 modules are IP/geo blocked, not bot-walled

`annaabi`, `donatello`, `producthunt` and `yaga_co_za` are untouched on purpose. Yaga's CloudFront response says it outright:

```
The Amazon CloudFront distribution is configured to block access from your country.
```

The other three answer *every* path, including `robots.txt`, with a challenge or firewall block, and a real stealth browser does not clear them either. I could not exercise a single code path on any of them, and an unverifiable transport swap could silently break them for contributors whose network is not blocked.

## Testing

32 real accounts and 33 invented-handle checks (3 controls × 11 modules), run against the live sites.

| Module | Real accounts | Invented handles | Metadata fields |
| --- | --- | --- | --- |
| academia | 3/3 Found | Not Found | 6-8 |
| adultism | 3/3 Found | Not Found | 13-15 |
| apclips | 3/3 Found | Not Found | 15-20 |
| kick | 2/2 Found | Not Found | 18 |
| npmjs | 3/3 Found | Not Found | 5-7 |
| osta | 3/3 Found | Not Found | 9-13 |
| reddit | 3/3 Found | Not Found | 14-16 |
| sourceforge | 3/3 Found | Not Found | 5-11 |
| themeforest | 3/3 Found | Not Found | 7-12 |
| tumblr | 3/3 Found | Not Found | 7-8 |
| unsplash | 3/3 Found | Not Found | 12-15 |

Edge cases:

```
reddit   awkwardtheturtle  Found     suspended, handle still owned
reddit   Deleted           Found     deleted, handle still owned
reddit   ab                Error     BAD_USERNAME, not retried into a verdict
tumblr   cyle              Found     login-walled blog
tumblr   PhotoMatt         Found     case normalised
osta     kristjan123       Found     case-resolved to Kristjan123
npmjs    babel             Found     org rather than user
npmjs    Isaacs            Error     uppercase rejected before the network
kick     XQC               Found     case-insensitive slug match
academia John.Smith        Error     period route is not addressable
```

Metadata was cross-checked against the live sources rather than trusted: Unsplash `nasa` 118 photos / 2 collections and Kick `xqc` 1,098,790 followers both match the APIs exactly.

A full 228-module scan shows no regression elsewhere — an invented handle went from 186 Not Found / 41 Error / **1 Found** to 195 / 33 / **0 Found**.

Concurrency: all 11 modules at `-C 20`, real and invented handle, no throttling errors. Reddit specifically was run 5 consecutive times at `-C 20` over 8 handles for **40/40 correct**, against roughly one failure per run before the retry.

**Not verified:**

- APClips fan-only accounts have no public page and no read-only check endpoint, so that type reports Not Found.
- Academia handles existing only on an institution subdomain (e.g. `unizg.academia.edu`) are out of scope, since those subdomains are unenumerable.
- Unsplash follower counts need an authenticated client_id and are omitted rather than substituted.
- Reddit accounts under a *temporary* suspension were not found to test.
- SourceForge's `availability` and `phone_numbers` formatters are coded to the Allura schema but no live profile populated them.
- Proxy mode and a non-default `-t` were not exercised.
- The Reddit retry will not rescue a sustained heavy scan — if most requests are shed, it returns `Error`, never a wrong verdict.

### DIFF
--- a/user_scanner/user_scan/adult/adultism.py
+++ b/user_scanner/user_scan/adult/adultism.py
@@ -1,146 +1,123 @@
+import html
 import re
-from user_scanner.core.orchestrator import generic_validate, Result, make_request
+
+from user_scanner.core.impersonate import impersonate_request, impersonate_validate
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.adultism.com"
 
 
-def validate_adultism(user):
-    url = f"https://www.adultism.com/profile/{user}"
-    show_url = f"https://www.adultism.com/profile/{user}"
+def validate_adultism(user: str) -> Result:
+    url = f"{BASE_URL}/profile/{user}"
 
     def process(response):
-        if response.status_code == 200:
-            is_taken = (
-                re.search(r'itemprop="name">\s*' + re.escape(user) + r'\s*</h1>', response.text, re.IGNORECASE) or
-                re.search(r'<title>\s*' + re.escape(user) + r'\s*-\s*Adultism</title>', response.text, re.IGNORECASE) or
-                re.search(r'title="' + re.escape(user) + r'"', response.text, re.IGNORECASE) or
-                re.search(r'alt="' + re.escape(user) + r'"',
-                          response.text, re.IGNORECASE)
-            )
-            if is_taken:
-                extra = {}
-                media = {}
+        if response.status_code == 404:
+            if "No such member" in response.text:
+                return Result.available(url=url)
+            return Result.error("Unexpected 404 (not the not-found page)", url=url)
 
-                # Extract Avatar
-                avatar_match = re.search(r'<img[^>]+src="([^"]+)"[^>]*class="[^"]*profile-image', response.text, re.IGNORECASE) or \
-                    re.search(
-                        r'<img[^>]+class="[^"]*profile-image[^"]*"[^>]*src="([^"]+)"', response.text, re.IGNORECASE)
-                if avatar_match and "defaults/member" not in avatar_match.group(1):
-                    media["image"] = avatar_match.group(1)
+        if response.status_code != 200:
+            return Result.error(f"Unexpected status: {response.status_code}", url=url)
 
-                # Extract Channel ID
-                channel_match = re.search(
-                    r'data-channel-id="(\d+)"', response.text, re.IGNORECASE)
-                if channel_match:
-                    extra["channel_id"] = channel_match.group(1)
+        profile = _profile_section(response.text)
+        if profile is None:
+            return Result.error("Profile confirmation not found", url=url)
 
-                # Extract Birth date
-                birthdate_match = re.search(
-                    r'<meta[^>]+itemprop="birthDate"[^>]+content="([^"]+)"', response.text, re.IGNORECASE)
-                if birthdate_match:
-                    extra["birthdate"] = birthdate_match.group(1)
+        # Lookups are case-insensitive and the heading carries the canonical
+        # spelling, so it is what ties the page to the requested handle.
+        nick = _nick(profile)
+        if nick.lower() != user.lower():
+            return Result.error("Profile heading does not match the handle", url=url)
 
-                # Extract Gender/Sex
-                sex_match = re.search(
-                    r'Sex:.*?<span[^>]*>\s*(.*?)\s*</span>', response.text, re.DOTALL | re.IGNORECASE)
-                if sex_match:
-                    cleaned_sex = re.sub(
-                        r"\s+", " ", sex_match.group(1)).replace("&nbsp;", " ").strip()
-                    extra["gender"] = cleaned_sex
-                else:
-                    gender_meta = re.search(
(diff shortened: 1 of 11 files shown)


kaifcodec (approved):
@brunolm It was tough reviewing all at once but I reviewed and tested every one of the modules and it's fine.

PR #523, opened 2026-08-06
fix(bentbox): detect profiles via the ProfilePage JSON-LD

brunolm:
> **This PR has been narrowed to BentBox only.** It kept its number because the discussion below concerns the conflict with #519. The other seven modules it used to carry are now #550–#556, listed at the bottom.

## tl;dr

- **Every handle errored**: the not-found branch keyed on a string that appears nowhere on the site, and the found gate keyed on boilerplate absent from the miss body.
- **Detection now keys on the ProfilePage JSON-LD**, which no other page on the site carries.
- **Only server-derived identity is reported** — the OpenGraph fields echo whatever casing you asked for.
- **Resolves the conflict with #519**, which is merged into `main`; this builds on top of it.

## Every handle errored

```
not-found branch keyed on:  "user is currently not available"   # appears nowhere on the site
real miss body:             "User not found"                    # 57 bytes
```

The found gate required site-wide boilerplate that the 57-byte miss body doesn't contain, so neither branch could fire and every handle fell through to an error.

## Detection keys on the ProfilePage JSON-LD

A hit is confirmed by the `ProfilePage` JSON-LD block, which no other page on the site emits, and metadata is read from its `Person` node.

## Only server-derived identity is reported

`og:title` and `profile:username` echo the **requested** casing, so they can't distinguish a real handle from the one you typed. Only the server-derived identifier and name are reported. The placeholder avatar served to creators without one is skipped rather than reported as their image.

## Relationship to #519

#519 is merged and this branch is rebased on top of it, so the earlier merge conflict is resolved and the diff here is the delta over #519's version. A follow-up commit also separates redirects from unrecognised 200 bodies — they previously shared one generic error — and drops the inherited message blaming Cloudflare for a 403, since this site runs no bot protection.

Routed through the impersonating transport, since httpx times out.

---

## Split out of this PR

| PR | Module |
| --- | --- |
| #550 | advfn |
| #551 | babepedia |
| #552 | bdsmlr |
| #553 | medium |
| #554 | stackoverflow |
| #555 | riot_id |
| #556 | bdsmsingles |

Every file in those PRs is byte-identical to what was tested here.

### DIFF
--- a/user_scanner/user_scan/adult/bentbox.py
+++ b/user_scanner/user_scan/adult/bentbox.py
@@ -1,106 +1,81 @@
+import json
 import re
 
 from user_scanner.core.orchestrator import Result, generic_validate
 
+# Served instead of an avatar when the creator uploaded none.
+PLACEHOLDER_AVATAR = "/img/Icon_BentBox"
+
 
 def validate_bentbox(user):
     url = f"https://bentbox.co/{user}"
-    show_url = f"https://bentbox.co/{user}"
+    show_url = url
 
     def process(response):
-        if "<h3>User not found</h3>" in response.text:
+        html = response.text
+
+        person = _profile_person(html)
+        if person:
+            return Result.taken(extra=_extra(person, html), media=_media(html), url=show_url)
+
+        if "User not found" in html:
             return Result.available(url=show_url)
-
-        if response.status_code == 200:
-            username_match = re.search(
-                r'"alternateName":\s*"([^"]+)"', response.text, re.IGNORECASE
-            ) or re.search(
-                r'<meta\s+property="profile:username"\s+content="([^"]+)"',
-                response.text,
-                re.IGNORECASE,
-            )
-            if (
-                username_match
-                and username_match.group(1).strip().casefold() == user.casefold()
-            ):
-
-                extra = {}
-                media = {}
-
-                # Extract profile data from OpenGraph or Title tags
-                title_match = re.search(
-                    r'<meta\s+property="og:title"\s+content="([^"]+)"', response.text, re.IGNORECASE)
-                if title_match:
-                    fullname = title_match.group(1).replace(
-                        "- BentBox Profile", "").replace("- BentBox", "").replace("on BentBox", "").strip()
-                    extra["fullname"] = re.sub(
-                        r'\(@[a-zA-Z0-9_-]+\)', '', fullname).strip()
-                else:
-                    t_tag = re.search(r"<title>(.*?)</title>",
-                                      response.text, re.IGNORECASE)
-                    if t_tag:
-                        fullname = t_tag.group(1).replace(
-                            "- BentBox Profile", "").replace("- BentBox", "").replace("on BentBox", "").strip()
-                        extra["fullname"] = re.sub(
-                            r'\(@[a-zA-Z0-9_-]+\)', '', fullname).strip()
-
-                # Extract Alternate Name / Username
-                extra["username"] = username_match.group(1).strip()
-
-                # Extract Bio
-                desc_match = re.search(
-                    r'<meta\s+property="og:description"\s+content="([^"]+)"', response.text, re.IGNORECASE)
-                if desc_match:
-                    extra["bio"] = desc_match.group(1).strip()
-
-                # Extract Avatar
-                image_match = re.search(
-                    r'<meta\s+property="og:image"\s+content="([^"]+)"', response.text, re.IGNORECASE)
-                if image_match:
-                    media["image"] = image_match.group(1).strip()
-
-                # Extract Identifier / User ID
-                id_match = re.search(
-                    r'"identifier":\s*"([^"]+)"', response.text, re.IGNORECASE)
-                if id_match:
-                    extra["profile_id"] = id_match.group(1).strip()
-
-                # Extract Followers
-                followers_match = re.search(
-                    r'(\d+)\s+Followers', response.text, re.IGNORECASE)
-                if not followers_match:
-                    follow_action = re.search(
-                        r'FollowAction".*?"userInteractionCount":\s*"(\d+)"', response.text, re.DOTALL | re.IGNORECASE)
-                    if follow_action:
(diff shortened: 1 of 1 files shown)


kaifcodec:
@kristoisberg We got a merge conflict here after the PR #519 merged.
Can you resolve the conflict testing which approach is better?

brunolm:
AI comparison

---

# Verdict

- **PR `#523` as it stands now: A+** - after the conflict resolution
- **PR `#523` as originally opened: A**
- **PR `#519`: C+** - it works, but it tells you three things that are not true

Both PRs only touch `bentbox.py`. The branch is rebased onto current main, the conflict is resolved, and the error messages have been rewritten beyond what either PR had.

## How to read the grades

| Grade | Meaning                                  |
| ----- | ---------------------------------------- |
| S     | Exactly right. Nothing to improve.       |
| A     | Right, with a small nitpick.             |
| B     | Works, but has a real weakness.          |
| C     | Works today, but it is fragile or messy. |
| D     | Works, but it reports something false.   |
| E     | Mostly broken.                           |
| F     | Completely broken.                       |

## Comparison

| Thing                      | `#519` | `#523` as opened | `#523` now |
| -------------------------- | ------ | ---------------- | ---------- |
| Fixes the original bug     | S      | S                | S          |
| Requests per scan          | S      | S                | S          |
| Transport choice           | S      | B                | S          |
| Not-Found marker           | B      | A                | A          |
| Found signal               | C      | A                | A          |
| Metadata: bio              | D      | S                | S          |
| Metadata: avatar           | D      | S                | S          |
| Metadata: `username` field | D      | S                | S          |
| Metadata: stats coverage   | C      | S                | S          |
| `media` key convention     | C      | S                | S          |
| Code structure             | C      | A                | A          |
| Error diagnostics          | D      | C                | S          |
| Comment accuracy           | n/a    | D                | S          |

## What the grades mean

- **Transport (S / B / S).** "Transport" is just which library fetches the page. `#519` uses the plain one (httpx). `#523` originally swapped in the heavier one that pretends to be Chrome, which you need on sites that block bots. BentBox does not block bots - I sent 104 requests, including bursts of 30 at once, and every one came back clean. So the heavy option cost something and bought nothing. That is now back to httpx, so all three agree.

- **Found signal (C vs A).** This is how each version decides "yes, this profile exists". `#519` searches the page for an `alternateName` tag and checks it matches what you typed. The catch is the site echoes back whatever you type, so that check can never fail - it is really just asking "does this tag exist". `#523` looks for the page's structured profile data (`ProfilePage` JSON-LD), which only real profiles carry. Much sturdier. It is A rather than S only because the site never exposes a canonical handle, so no version can truly confirm the profile belongs to the name you asked for.

- **The `username` field (D vs S).** `#519` fills this with the exact text you searched for. Ask for `SimonScans`, it says `username: SimonScans`. Ask for `simonscans`, it says `simonscans`. That is your own input handed back, dressed up as profile data. I checked on the wire: `og:title`, `og:url`, `profile:username` and `alternateName` all mirror your typing, and only the structured `Person.name` stays fixed at `Simonscans`. `#523` drops the field rather than print something misleading.

- **Avatar (D vs S).** When an account never uploaded a picture, the site serves its own logo. `#519` reports that logo as the person's face:
  ```
  admin
    519  media={'image': 'https://bentbox.co/img/Icon_BentBox_Nov_2025.jpg'}
    523  media={}
  ```

- **Bio (D vs S).** `#519` reads the short preview text the site truncates around 155 characters, so you get `"...and also the very cre..."` with the rest gone. `#523` reads the real bio, all 580 characters.

- **Stats (C vs S).** `#519` hardcodes two numbers, boxes and videos. `#523` reads whatever counters the page shows and picked up `collections: 108`, which `#519` never sees.

- **`media` key name (C vs S).** The rest of the codebase calls the profile picture `avatar` (70 places). `#519` calls it `image` (14 places). `#523` matches the majority.

- **Code structure (C vs A).** `#519` is 106 lines inside one deeply nested function. `#523` splits it into three small named helpers.

- **Error diagnostics (D / C / S).** Neither PR was good here, so it got rewritten. `#519` carries a message inherited word for word from the original broken module, blaming Cloudflare and telling you to get a residential proxy - for a site with no Cloudflare and no bot protection at all. `#523` removed it but left one vague catch-all. The current version names what actually happened:
  ```
  login    -> Redirected to https://bentbox.co/signin; the response carries no verdict
  terms    -> Neither a profile nor the not-found marker (HTTP 200)
  ```

- **Comment accuracy (D vs S).** `#523` originally had a comment claiming handles with dashes or spaces redirect to `/page-not-available`. Not true - `kink-artist` returns a normal "not found" with no redirect. It is gone, replaced by one describing the two redirects that do happen: reserved routes like `/login` and `/explore`, and the server tidying spaces in the URL. The three remaining comments were each checked against live responses.

## Summary

`#519` is a real fix and it stops the errors. But it gives you a truncated bio, the site logo as someone's face, and your own search text labelled as profile data. `#523` gets the same verdicts and is honest about what it knows. After the conflict resolution it also picks up `#519`'s one genuine advantage, the lighter transport, so nothing from `#519` is lost.

brunolm:
Narrowed this PR to **BentBox only** and split the other seven modules into their own PRs, so each can be reviewed and merged independently. Kept the number here because the #519 conflict discussion lives in this thread.

- #550 advfn · #551 babepedia · #552 bdsmlr · #553 medium
- #554 stackoverflow · #555 riot_id · #556 bdsmsingles

@kaifcodec on the conflict: #519 is merged into `main` and this branch is now rebased on top of it, so the conflict is resolved and the diff here is purely the delta over #519's version.

No code changed in the split — every file across the eight PRs is byte-identical to what was tested here, and `ruff`/`mypy` pass on each branch. Force-pushed to reduce the diff; there were no inline review comments to orphan.

kristoisberg:
@kaifcodec I didn't run any tests, but looking at it, everything seems good. JSON-LD is much easier to parse and all known fields are covered in a more future-proof way.

kaifcodec (approved):
LGTM. I tested it with known and invented usernames!

PR #524, opened 2026-08-06
fix: unblock apexlegends, buzzfeed, cups7 and flickr

brunolm:
## tl;dr

- **Buzzfeed and Flickr reported reserved site pages as accounts.** `/quizzes`, `/tasty`, `/photos/search` and `/photos/me` all soft-200; both modules now require the payload to name the requested handle.
- **Apexlegends checked one of three account namespaces.** Apex accounts are per-platform, so a PSN-only account was reported free. Now checks `origin`, `psn` and `xbl`.
- **Apexlegends, Buzzfeed and Flickr were fingerprint-blocking httpx** (403 / 406 / 502), returning errors for every handle. Moved to curl_cffi.
- **7 Cups profile pages are behind an AWS WAF JS challenge** no HTTP client can clear. Replaced with the unchallenged JSON endpoint that backs the same page — which also takes the module from zero `extra` fields to twenty.
- **Apexlegends now costs 3 requests per handle**, so tracker.gg's rate limit arrives 3x sooner. Blocked legs return `error`, never a verdict.

Handles below are written as `<real>` / `<real-listener>` etc. rather than the live accounts they were tested against.

## Buzzfeed and Flickr reported reserved site pages as accounts

Both trusted a bare 200. Buzzfeed fell through to `Result.taken()` whenever `__NEXT_DATA__` had no user node; Flickr returned `taken` for any 200 at all.

```
buzzfeed/quizzes  -> TAKEN      # section page
flickr/search     -> TAKEN      # site page under /photos/
flickr/me         -> TAKEN
```

Buzzfeed now requires `props.pageProps.user`, and Flickr requires the photostream owner's `pathAlias` or NSID to match the requested handle. Both return `error` on a 200 that carries neither — the name is not free, but nothing confirms an account.

Flickr's owner match is case-insensitive and accepts the NSID, since handles resolve at any casing and accounts with no custom URL are addressed as `12345678@N00`:

```
flickr/<Real>            -> FOUND   # canonical path alias is lowercase
flickr/<nsid>@N01        -> FOUND   # same account
```

## Apexlegends checked one of three account namespaces

Only `origin` was queried, so a PSN- or Xbox-only account was reported available.

| Handle | origin | psn | xbl |
| --- | --- | --- | --- |
| `<real-multi-platform>` | 200 | 200 | 200 |
| `<real-psn-only>` | 404 | **200** | 404 |
| `zzznotarealuser99xzq` | 404 | 404 | 404 |

`available` now needs a clean 404 on all three; if any leg was blocked, that error is returned instead. When more than one platform hits, `extra` keys are prefixed (`origin_level`, `psn_kills`) and `platforms` lists them.

### The psn/xbl legs are not stubs or catch-alls

Some psn/xbl hits come back thin (level 2-3, 0 kills), which looks like a placeholder. Three checks say they are real accounts:

| Probe | Result |
| --- | --- |
| 5 invented but plausible gamertags | 404 on psn **and** xbl — no profile is created on lookup |
| 8 confirmed-real PSN IDs sampled from a trophy leaderboard | 7 x 404, 1 x 200 — not a PSN-existence catch-all |
| 3 tracker.gg xbl hits vs. an independent gamertag directory | all 3 exist on Xbox Live; the invented control is absent from both |

The single leaderboard ID that hit came back `pageviews=0` on a first-ever request, so the profile was resolved live from EA rather than left over from earlier traffic. A thin profile is someone who launched Apex on that platform and stopped.

One consequence worth stating: a psn/xbl 404 means no *Apex* account on that platform, not that the gamertag is unclaimed there.

## Apexlegends, Buzzfeed and Flickr were fingerprint-blocking httpx

| Module | httpx | curl_cffi |
| --- | --- | --- |
| api.tracker.gg | 403 every request | 200 / 404 |
| flickr.com | 502 every request | 200 / 404 |
| buzzfeed.com | 406 on cache miss | 200 / 404 |

Buzzfeed's was the intermittent one, and it is a CDN artifact — the edge serves a cached response to anyone, so httpx only fails when it misses. Appending a cache-busting query makes it deterministic:

```
httpx  /<handle>        -> 404 404 404      # edge cache hit
httpx  /<handle>?i=N    -> 406 406 406      # cache miss, origin sees httpx
cffi   /<handle>?i=N    -> 404 404 404
```

## 7 Cups profile pages are behind an AWS WAF JS challenge

`/@handle` answers 202 with `x-amzn-waf-action: challenge` for real and fake handles alike, under both transports. So do `/forum/`, `/qa/` and the forum sitemap; the homepage and `/community/` are unprotected.

`/apiv2/user/<name>` backs the same profile page, is not challenged, and needs no auth:

```
/apiv2/user/<real-listener>      -> 200  {"listID":"...","listType":"listener",...}
/apiv2/user/<real-member>        -> 200  {"memID":"...","memberLevel":"...",...}
/apiv2/user/zzznotarealuser99xzq -> 404  {}
```

It also distinguishes the two account types, which the module now reports: listeners carry `listID`/`listType`, plain members carry `memID`. `extra` goes from empty to `screen_name`, `type`, `user_id`, `bio`, `quote`, `age`, `country`, `language`, `joined`, `listener_since`, `level`, `points`, `conversations`, `compassion_hearts`, `forum_upvotes`, `rating`, `last_active`, `online`, `badges` and an avatar.

`robots.txt` disallows `/apiv2/`. This is a single targeted lookup rather than crawling, and it is the only path that yields a verdict at all, but it is a deliberate choice worth naming — the alternative is a module that always errors.

## Apexlegends now costs 3 requests per handle

tracker.gg is rate-limited at the Cloudflare edge and answers 429 with `retry-after: 44` after roughly 15 lookups. Tripling the per-handle request count brings that on three times sooner during bulk scans. A 429 returns `Result.error`, so it never becomes a false verdict — but a large `-uf` run will see more Apex errors than before.

## Before / after

| Module | Handle | Before | After |
| --- | --- | --- | --- |
| Apexlegends | `<real-psn-only>` | Error (403) | Found — PlayStation only |
| Apexlegends | `<real-multi-platform>` | Error (403) | Found — EA + PSN + Xbox |
| Buzzfeed | `<real-contributor>` | Error (406) | Found — name, bio, join date, post count |
| Buzzfeed | `quizzes` | Found | Error (no profile data) |
| Cups7 | `<real-listener>` | Error (WAF challenge) | Found — listener, rating, conversation count |
| Cups7 | `<real-member>` | Error (WAF challenge) | Found — member, level, join date |
| Flickr | `<real>` | Error (502) | Found — photo, follower and following counts |
| Flickr | `search`, `me` | Found | Error (no matching owner) |
| all four | `zzznotarealuser99xzq` | Error / Found | Available |

## Testing

Live-tested against real and nonexistent handles on every namespace each module checks, plus one handle per account type: Apex on origin / psn / xbl, 7 Cups listener and member, Flickr path-alias and NSID, Buzzfeed contributor and section pages. `extra` was read back against the live pages.

Not verified:

- **7 Cups therapist accounts.** `userType` `"t"` is mapped to `therapist` by inference from the `"l"` / `"m"` pattern; no confirmed-real therapist handle was available to test.
- **7 Cups invalid handles.** There is no name-check endpoint, so a too-short name returns the same 404 as a free one and is reported available.
- **Apex profiles that are neither thin nor a pro's.** The psn/xbl legs were checked against invented handles, real-but-unrelated PSN IDs and an independent Xbox directory (above); mid-range accounts were not sampled specifically.

### DIFF
--- a/user_scanner/user_scan/gaming/apexlegends.py
+++ b/user_scanner/user_scan/gaming/apexlegends.py
@@ -1,53 +1,99 @@
-from user_scanner.core.orchestrator import generic_validate, Result
+from user_scanner.core.impersonate import impersonate_request
+from user_scanner.core.result import Result
 
-def validate_apexlegends(user):
-    url = f"https://api.tracker.gg/api/v2/apex/standard/profile/origin/{user}"
-    show_url = f"https://apex.tracker.gg/apex/profile/origin/{user}/overview"
+API_URL = "https://api.tracker.gg/api/v2/apex/standard/profile"
+PROFILE_URL = "https://apex.tracker.gg/apex/profile"
 
-    headers = {
-        "Accept-Language": "en-US,en;q=0.5",
-        "Origin": "https://apex.tracker.gg",
-        "Referer": "https://apex.tracker.gg/",
-        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:128.0) Gecko/20100101 Firefox/128.0",
-    }
+# Apex accounts are per-platform: a handle can resolve on PSN and miss on the
+# other two, so a single namespace never settles availability.
+PLATFORMS = {"origin": "EA", "psn": "PlayStation", "xbl": "Xbox"}
 
-    def process(response):
-        if response.status_code == 200:
-            try:
-                payload = response.json()
-                data = payload.get("data", {})
-                extra = {}
-                media = {}
+HEADERS = {
+    "Accept-Language": "en-US,en;q=0.5",
+    "Origin": "https://apex.tracker.gg",
+    "Referer": "https://apex.tracker.gg/",
+}
 
-                p_info = data.get("platformInfo", {})
-                if handle := p_info.get("platformUserHandle"): extra["username"] = handle
-                if avatar := p_info.get("avatarUrl"): media["avatar"] = avatar
 
-                u_info = data.get("userInfo", {})
-                if country := u_info.get("countryCode"): extra["country"] = country
-                if u_info.get("isPremium"): extra["premium"] = "Yes"
-                if u_info.get("isInfluencer"): extra["influencer"] = "Yes"
+def validate_apexlegends(user: str) -> Result:
+    show_url = f"{PROFILE_URL}/origin/{user}/overview"
 
-                meta = data.get("metadata", {})
-                if legend := meta.get("activeLegendName"): extra["active_legend"] = legend
+    hits: dict[str, dict] = {}
+    errors: list[str] = []
 
-                segments = data.get("segments", [])
-                if segments:
-                    stats = segments[0].get("stats", {})
-                    if lvl := stats.get("level"): extra["level"] = str(lvl.get("displayValue"))
-                    if kills := stats.get("kills"): extra["kills"] = str(kills.get("displayValue"))
-
-                return Result.taken(extra=extra, media=media)
-            except Exception:
-                pass
-            return Result.error("200 response status with no recognizable data, report it via GitHub issues")
+    for platform in PLATFORMS:
+        try:
+            response = impersonate_request(
+                f"{API_URL}/{platform}/{user}", headers=HEADERS
+            )
+        except Exception as e:
+            errors.append(f"{platform}: {e}")
+            continue
 
         if response.status_code == 404:
-            return Result.available()
+            continue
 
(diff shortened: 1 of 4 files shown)


kaifcodec (approved):
LGTM.

PR #525, opened 2026-08-06
chore: retire Motherless modules to abandoned/ (site is offline)

brunolm:
Retires both Motherless modules — `user_scan/adult/motherless.py` and `email_scan/adult/motherless.py` — by moving them to `abandoned/<email_scan|user_scan>/adult/`, unchanged. The site no longer exists, so both modules could only ever return `Result.error` after burning a full timeout.

`motherless.com` was suspended at the Verisign registry in June 2026 under a Texas age-verification order; the site moved to `motherless.xxx`, which is what these modules target. That domain no longer resolves either:

```
motherless.xxx -> Non-existent domain   (8.8.8.8 and 1.1.1.1)
motherless.com -> Non-existent domain   (8.8.8.8 and 1.1.1.1)
```

Dutch police seized the site's servers on 21 July 2026 (reported by NL Times / CNN), so this is not a transient outage or a geo-block — there is no A record to reach from anywhere.

Also documents the retirement convention in `AGENTS.md` (never delete a module; move it to `abandoned/` preserving scan type, category, and file name) so future removals follow it.

## Verification

Username scan before retiring, from my location:

```
brunolm7             -> Error: ConnectError: Could not resolve hostname
zzznotarealuser99xzq -> Error: ConnectError: Could not resolve hostname
```

After the move, `-m motherless` reports `User module 'motherless' not found.`, confirming `load_modules()` does not pick up `abandoned/`.

No other tracked file references the site (`git grep -i motherless` outside `abandoned/` is clean), so nothing else needed updating — there is no site list or registry to prune.


kaifcodec:
@brunolm We don't remove the modules that are abandoned, we move them to `abandoned/<email_or_user>/<respective_category>/...` can you move these there?

kaifcodec (approved):
That AGENTS.md update is clean.
LGTM.

PR #526, opened 2026-08-07
fix(user_scan): restore 5 modules that could not return a verdict, report 4 blocked ones honestly

brunolm:
## tl;dr

- **Five username modules could not return a verdict at all; they can now.** SpankBang, Facebook, LiveJournal, DefensiveCarry and TheFirearmsForum each errored on every handle. All five are live-verified against a real and an invented handle.
- **Facebook had two verdict bugs beyond the transport.** Non-canonical casing 302s were never followed, and the not-found marker has a second wording the exact-string check missed.
- **DefensiveCarry and TheFirearmsForum inferred availability from a bare 200.** They now require the explicit "member cannot be found" marker.
- **LiveJournal usernames treat `-` and `_` as one character.** The username guard normalises both sides, or `james-nicoll` would error while `james_nicoll` resolves.
- **Purged and suspended LiveJournal accounts get opposite verdicts, and both used to error.** A purged journal frees its username; a suspended one keeps it.
- **Four modules are unreachable behind Cloudflare/CloudFront walls and stay that way.** They now name the wall instead of emitting a bare 403. Annaabi's verdict logic was verified through a real browser; the other three could not be exercised at all.
- **Product Hunt still returns `taken` on any 200.** Deliberate: the site is challenge-walled from here, so a stricter marker could not be verified.

## Five username modules could not return a verdict at all; they can now

| Module | Before | After (real handle) | After (invented handle) |
| --- | --- | --- | --- |
| `adult/spankbang.py` | `Unexpected status: 403` | `brunolm7` → Found | Not Found |
| `social/facebook.py` | `Unexpected response body` | `zuck`, `cristiano`, `bill.gates` → Found | Not Found |
| `community/defensivecarry.py` | `ReadTimeout` | `admin`, `Jeff` → Found | Not Found |
| `community/thefirearmsforum.py` | `ReadTimeout` | `admin`, `Bob` → Found | Not Found |
| `social/livejournal.py` | TLS handshake timeout | `ohnotheydidnt`, `james_nicoll` → Found | Not Found |

What each needed:

- **SpankBang** — Cloudflare now challenges the *Chrome* fingerprint on every path of the domain. `safari17_0` is let through; every `chrome*` and `edge*` target returns `cf-mitigated: challenge`.
- **Facebook** — urllib got a body with neither marker; curl_cffi gets the real page.
- **The two XenForo forums** — the proof-of-work solver was already correct, httpx just stalled on it. Over curl_cffi the full solve-and-replay runs in under a second.
- **LiveJournal** — the journal subdomain serves ~750 KB and stalls Python's TLS stack. `www.livejournal.com/profile/?user=` carries the same `Site.journal` payload.

All five now populate `extra` — numeric ids, display names, account type, member counts:

```
spankbang  mia_khalifa   {'name': 'mia_khalifa', 'user_id': '51279794'}
livejournal ohnotheydidnt {'uid': 3616053, 'name': 'ohnotheydidnt', 'type': 'community', 'is_paid': True, ...}
facebook   bill.gates    {'name': 'Bill Gates', 'handle': 'BillGates', 'likes': '41,470,070', ...}
```

## Facebook had two verdict bugs beyond the transport

| Fault | Effect |
| --- | --- |
| 302 to canonical casing never followed | `bill.gates` (→ `BillGates`) missed entirely |
| Marker matched as one exact string | Facebook also ships "This content isn't available **right now**", apostrophe backslash-escaped inside JSON |

The marker is now a regex covering both wordings and the escaped form; the canonical `og:url` also supplies the handle's real casing.

## DefensiveCarry and TheFirearmsForum inferred availability from a bare 200

```python
elif response.status_code == 200 or response.status_code == 404:
    return Result.available(url=url)   # fires for the challenge page too
```

A miss is now confirmed on `"The specified member cannot be found"`, and a hit is confirmed by a redirect matching `/members/<slug>.<id>/` — an unrecognised redirect target returns an error rather than a hit. The member id and canonical slug come out of that redirect, so no extra request is needed.

The two files stay near-identical copies rather than sharing a helper: `load_modules()` globs every `*.py` under `user_scan/<category>/` and would load a shared helper as a site module with no `validate_` function.

## LiveJournal usernames treat `-` and `_` as one character

```
james-nicoll -> TAKEN (uid 86776258)   # same journal
james_nicoll -> TAKEN (uid 86776258)
JAMES_NICOLL -> TAKEN (uid 86776258)
```

The rewrite guards against the generic profile shell by checking the echoed `display_username`, so that comparison normalises case and `-`/`_` on both sides. Also removed: the old `302/301 → available` branch, which turned every redirect into "this name is free". The old `403 → taken (suspended)` branch was right for this site and is kept, but gated on an explicit marker instead of the bare status — see below.

## Purged and suspended LiveJournal accounts get opposite verdicts

```
deleted -> 410 "Purged Account"    "This journal has been deleted and purged.
                                    You can rename your account with this username."  -> AVAILABLE
abc     -> 403 "Suspended Journal" "This journal has been suspended."                 -> TAKEN
```

Both statuses fell through to `Unexpected status` before. The distinction is not cosmetic — a purged username is genuinely re-registrable, so calling it taken would be wrong in the other direction.

## Four modules are unreachable behind Cloudflare/CloudFront walls

| Module | Wall | Reported now |
| --- | --- | --- |
| `creator/producthunt.py` | Cloudflare managed challenge, whole domain | `Cloudflare challenge, cannot be solved without a browser` |
| `learning/annaabi.py` | Cloudflare managed challenge | same |
| `donation/donatello.py` | Cloudflare regional block ("Attention Required") | `Blocked by Cloudflare, the site is unreachable from here` |
| `shopping/yaga_co_za.py` | CloudFront: "configured to block access from your country" | `CloudFront blocks this country, the site is unreachable from here` |

These are network-level walls at my location, not module logic — no HTTP client clears a managed challenge, and every impersonation target returns the same `cf-mitigated: challenge`. Annaabi is the one case where that claim is proven rather than assumed: through a real browser its check endpoint answers normally and both of its markers still match (see the testing section). The first three moved to the impersonating transport anyway, which may be all they need where the site is reachable. `shopping/yaga_ee.py` got the same geo-block detection purely to keep the sibling pair identical; it is not defective.

## Product Hunt still returns `taken` on any 200

Deliberate, not an oversight. The module is rewritten onto the impersonating transport here, but the domain is challenge-walled from my location, so I could not see a real profile page or a miss page — adding a marker check would mean guessing at markup and risking a false `error` on every hit. The pre-existing soft-200 risk is left as it was, and flagged here rather than changed blind.

## Testing, and what was not tested

Live-tested against a real and an invented handle for every module that responds, plus a full 228-module scan before and after on both handles to check for collateral damage. No module moved in the wrong direction; the only regressions in that comparison (`fiverr`, `lastfm`, `nexusmods`, `discord`) are in files this branch does not touch and flip between consecutive runs on identical code.

The four walled sites were re-checked through a real browser, which separates "the module logic is wrong" from "the site will not talk to this location":

| Module | Real browser result | Verdict logic |
| --- | --- | --- |
| `learning/annaabi.py` | Reachable | Verified — `admin`, `test`, `marko` return the exact taken marker; invented handles the exact free marker |
| `donation/donatello.py` | Cloudflare block page | Unverifiable from here |
| `shopping/yaga_co_za.py` | CloudFront country block | Unverifiable from here |
| `creator/producthunt.py` | Challenge loops, never clears | Unverifiable from here by any means |

Remaining gaps:

- **Product Hunt's soft-200 risk stays untested** — the challenge does not clear in a real browser either, so neither a profile page nor a miss page is observable from this location.
- **LiveJournal connectivity is intermittent here** — 4 connection timeouts in 30 consecutive requests. Those surface as `Result.error`, never a verdict.
- **LiveJournal deleted-but-not-yet-purged journals** — purged (410) and suspended (403) are covered by real handles; the window between deletion and purge was not observed and may behave differently.
- **Facebook alias handles resolve to their canonical account** — `facebook.com/mark` redirects to Mark Zuckerberg's profile, so `mark` reports taken with `handle: zuck`. Accurate (the handle is not free), but `extra` names a different string than the one scanned.
- **Facebook profiles with no like count** — verified rather than assumed: `jack` returns `{'name': 'Jack Lindamood', 'handle': 'jack'}` and nothing else.

### DIFF
--- a/user_scanner/user_scan/adult/spankbang.py
+++ b/user_scanner/user_scan/adult/spankbang.py
@@ -5,11 +5,19 @@ from user_scanner.core.impersonate import impersonate_validate
 from user_scanner.core.result import Result
 
 
+# Cloudflare serves a managed challenge to the Chrome fingerprint on every
+# path of this domain; the Firefox one is let through.
+IMPERSONATE = "firefox133"
+
+
 def validate_spankbang(user):
     url = f"https://spankbang.com/profile/{user}"
     show_url = url
 
     def process(response):
+        if response.headers.get("cf-mitigated") == "challenge":
+            return Result.error("Cloudflare challenge, cannot be solved without a browser", url=show_url)
+
         if response.status_code == 404:
             return Result.available(url=show_url)
 
@@ -24,7 +32,7 @@ def validate_spankbang(user):
 
         return Result.error("Profile confirmation not found", url=show_url)
 
-    return impersonate_validate(url, process, show_url=show_url)
+    return impersonate_validate(url, process, impersonate=IMPERSONATE, show_url=show_url)
 
 
 def _is_profile(html_text: str, user: str) -> bool:
--- a/user_scanner/user_scan/adult/spankbang.py
+++ b/user_scanner/user_scan/adult/spankbang.py
@@ -6,8 +6,8 @@ from user_scanner.core.result import Result
 
 
 # Cloudflare serves a managed challenge to the Chrome fingerprint on every
-# path of this domain; the Firefox one is let through.
-IMPERSONATE = "firefox133"
+# path of this domain; the Safari one is let through.
+IMPERSONATE = "safari17_0"
 
 
 def validate_spankbang(user):
--- a/user_scanner/user_scan/community/defensivecarry.py
+++ b/user_scanner/user_scan/community/defensivecarry.py
@@ -1,73 +1,78 @@
 import hashlib
 import re
-from user_scanner.core.helpers import get_random_user_agent
-from user_scanner.core.orchestrator import Result, make_request
+from typing import Optional
+
+from user_scanner.core.impersonate import impersonate_request
+from user_scanner.core.result import Result
+
+BASE_URL = "https://www.defensivecarry.com"
+
+CHALLENGE_RE = re.compile(r"(\w+):'([^']*)'")
+CHALLENGE_KEYS = (
+    "challenge_nonce",
+    "challenge_hmac",
+    "difficulty",
+    "difficulty_char",
+    "issued_at",
+)
+NOT_FOUND_MARKER = "The specified member cannot be found"
+PROFILE_PATH_RE = re.compile(r"/members/([^/]+)\.(\d+)/?$")
+
 
 def validate_defensivecarry(user: str) -> Result:
-    url = f"https://www.defensivecarry.com/members/?username={user}"
-    
-    headers = {
-        "User-Agent": get_random_user_agent(),
-        "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8",
-        "Accept-Language": "en-US,en;q=0.9",
-    }
-    
+    url = f"{BASE_URL}/members/?username={user}"
+
     try:
-        response = make_request(url, follow_redirects=False, http2=False, headers=headers)
-        
+        response = impersonate_request(url)
+
+        # The site answers 202 with a JavaScript proof-of-work page; solving it
+        # and replaying with the pow_bypass cookie returns the real response.
         if response.status_code == 202:
-            html_text = response.text
-            nonce_match = re.search(r"challenge_nonce:'([^']+)'", html_text)
-            hmac_match = re.search(r"challenge_hmac:'([^']+)'", html_text)
-            diff_match = re.search(r"difficulty:'([^']+)'", html_text)
-            char_match = re.search(r"difficulty_char:'([^']+)'", html_text)
-            issued_match = re.search(r"issued_at:'([^']+)'", html_text)
-            
-            if (
-                nonce_match is None or
-                hmac_match is None or
-                diff_match is None or
-                char_match is None or
-                issued_match is None
-            ):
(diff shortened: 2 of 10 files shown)


brunolm:
Superseded — split into one PR per module so each can be reviewed and merged independently.

**Restored: modules that could not return a verdict**
- #541 — spankbang: impersonate Safari to clear the Cloudflare challenge
- #542 — facebook: follow canonical-casing redirects, match both miss wordings
- #543 — livejournal: resolve journals, split purged from suspended
- #544 — defensivecarry: confirm a miss instead of inferring it from a bare 200
- #545 — thefirearmsforum: same fault, same fix

**Walled sites: name the wall instead of emitting a bare 403**
- #546 — producthunt (soft-200 risk deliberately left, see the PR)
- #547 — annaabi (verdict logic verified through a real browser)
- #548 — donatello
- #549 — yaga (both siblings, reasoning in the PR)

No code changed in the split: every file is byte-identical to what was tested here, coverage of this branch's ten files is complete, and `ruff` and `mypy` pass on all nine branches. The branches are rebased onto current `main` (which now includes #525).

PR #527, opened 2026-08-07
fix(email_scan): restore 11 modules broken by parser rot and bot walls

brunolm:
## tl;dr

- **Eleven email modules returned `Error` for every address, and five of them were sitting on accounts they could already see.** WordPress, NYTimes, Patreon, Pornhub and GitHub all had the answer in hand and threw it away.
- **Pornhub now leans on an enumeration leak that Pornhub can close at any time.** The check no longer discloses registration for the address as typed; the module probes a sub-addressed alias instead.
- **A GitHub miss no longer means "available".** The search API only sees addresses published on a public profile, so the module says so rather than returning Not Registered.
- **`core/impersonate.py` gains a transport bridge shared by both scan types.** Eight modules now route through it; a change there affects username scans too.
- **Five modules from the same report are deliberately untouched.** They are blocked by challenges and geo walls no HTTP client can clear.

## Eleven email modules returned `Error` for every address

Verified against a real address and an invented one. Real addresses are redacted; `test@gmail.com` stands in where a widely-registered address was needed to exercise the taken branch.

| Module | Fault | Before | After |
| --- | --- | --- | --- |
| wordpress | `403 email_login_not_allowed` **is** the exists signal, thrown as an error | `WordPress Error: Please log in using your username…` | Registered + `login_method: username only` |
| deezer | Keyed on a `country` error the API stopped returning, so nothing matched | `Unexpected response body structure` | verdict |
| nytimes | New `show-welcome-back` action for SSO-only accounts fell through to `else` | `Got an weird action: show-welcome-back` | Registered + `sso_provider: google` |
| patreon | New `sso_required` auth step, on top of a 403 | `Status 403` | Registered + `sso_providers` |
| kick | httpx fingerprint-blocked | `Caught by Cloudflare WAF (403)` | verdict |
| fapfolder | httpx fingerprint-blocked | `403` | verdict |
| letterboxd | Cloudflare served the challenge page where JSON was expected | `JSONDecodeError` | verdict |
| walmart | httpx fingerprint-blocked | `412 session mismatch` | verdict |
| classmates | httpx fingerprint-blocked | `WAF (403) during Handshake` | verdict |
| pornhub | Token relocated; homepage now 410s to httpx | `Failed to extract dynamic token` | Registered |
| github | `/signup` behind a DataDome interstitial | `Failed to extract authenticity_token` | Registered + login/id/profile/avatar |

Two of the transport swaps needed more than the swap. Walmart and classmates were re-tripping their own WAF because the module's hand-set `User-Agent` and client hints contradicted the impersonated TLS fingerprint — dropping them fixed both. Walmart also cannot use the shared `warmup_url`, which warms once per session key: its warm-up URL carries a per-call PKCE challenge, so it fetches explicitly.

## Pornhub now leans on an enumeration leak

For the address as typed, the check is deliberately uninformative — this is the same string for a registered address and for two that plainly are not:

```
<registered>@gmail.com -> "If this email is already registered, you'll receive a verification link."
support@github.com     -> "If this email is already registered, you'll receive a verification link."
postmaster@gmail.com   -> "If this email is already registered, you'll receive a verification link."
```

Pornhub strips `+tag` sub-addressing before the lookup, and forgot to apply the placeholder on that path:

```
<registered>+zz1@gmail.com   -> {"email":"create_account_failed","error_message":"Email has been taken."}
<unregistered>+zz1@gmail.com -> {"email":"create_account_passed","error_message":""}
```

Nobody registered `<registered>+zz1@gmail.com`, so "taken" can only be about the base address after normalisation. That is what makes the probe read the real account, and it is also why this is the most fragile thing in the diff: it is an undocumented inconsistency, not an API contract. If Pornhub applies the placeholder uniformly, the module returns `Unexpected API response` rather than a wrong verdict.

The probe needs the alias to be deliverable, so a dead mailbox or a provider without sub-addressing yields `Email is invalid or cannot be used` — reported as an error, never a guess. Addresses already containing `+` are rejected up front.

It also avoids the "you'll receive a verification link" path entirely, which is the only one that could plausibly mail the account holder.

## A GitHub miss no longer means "available"

`/signup` returns a DataDome interstitial on every impersonation target tried (`chrome`, `chrome131`, `chrome136`, `firefox135`, `safari184`, `edge101`), with and without the client hints the response asks for. The authenticity_token is unreachable, so the module leads with the unauthenticated search API:

```
GET https://api.github.com/search/users?q=<email>+in:email
```

A hit is definitive and carries `login`, `id`, `html_url`, `type` and the avatar. A miss is not — the API only sees addresses published on a profile — so the module falls through to the signup probe and, when that is blocked, returns an error naming both reasons instead of a verdict.

Search has its own rate-limit bucket, so this does not compete with `githubgist` for the 60/hr core-API budget.

## `core/impersonate.py` gains a transport bridge shared by both scan types

Every `email_scan` validator is `async def`; curl_cffi's session API blocks. `_run_batch` fans out up to 25 modules on one event loop, so a blocking call there stalls every module in flight.

`impersonate_request_async` is nine lines delegating to the existing `impersonate_request` via `asyncio.to_thread`, so session caching, cookie warm-up, proxy rotation and the `-t` timeout stay in one place. The consequence worth knowing: the curl_cffi session cache is now shared across email and username scans rather than username only.

The alternative is inlining `asyncio.to_thread(impersonate_request, …)` at each of the eight call sites — identical behavior, no core diff, ~10 duplicated lines. Happy to switch if the shared cache is not wanted.

## Five modules from the same report are deliberately untouched

Not oversights — each was investigated and found unreachable, so shipping churn against them would only add noise:

| Module | Wall |
| --- | --- |
| render | Non-enumerable by design. `signUp` answers `{"email":"invalid"}` for everything including the literal string `notanemail`; a junk `hcaptchaToken` short-circuits to `{"hcaptcha_token":"invalid"}` before email is evaluated; `signIn` returns one unified "incorrect or this account doesn't exist" for real and fake. No account was created while establishing this. |
| annaabi | Cloudflare managed challenge site-wide, homepage included; unchanged across httpx, curl_cffi and urllib |
| nykaaman | Akamai edge 403 on the whole host, homepage included |
| emirates | Akamai bot manager resets the stream even with warm `_abck`/`bm_sz` cookies |
| nextdoor | Nextdoor's own nameservers answer `0.0.0.0`, so the host never resolves |

`tube8` is also excluded: it fails on rate limiting, and probing it enough to fix it would mean tripping that limit repeatedly.

## Testing

Live, against a real address and an invented one, on all eleven modules.

```
<registered>@gmail.com          -> 6 Registered, 5 Not Registered, 0 Error
zzznotarealuser99xzq@gmail.com  -> 0 Registered, 0 false positives
```

Taken branches exercised separately via `test@gmail.com`, which flipped letterboxd, deezer, nytimes, wordpress, fapfolder, patreon, kick and walmart to Registered — so both branches are live-verified for those eight, plus pornhub and github from the real address.

### Verified after review

Four gaps flagged in the first revision have since been closed, and two of them turned up bugs now fixed in `2d993ce`.

| Gap | Result |
| --- | --- |
| classmates' taken branch | **Verified.** A registered address returns `invalid registration/password`; the module reports Registered. Note Classmates rate-limits after ~4 rapid attempts, which the module already handles as a 429. |
| github's available branch | **Verified via a real browser**, which clears the DataDome interstitial: `200` + `Email is available` for an unregistered address, `422` + `already associated with an account` for a registered one. |
| pornhub on non-gmail providers | **Verified** on outlook, hotmail, yahoo, icloud, gmx, yandex and fastmail. A nonsense local part returns `create_account_passed` at every one of them, with and without the tag, so the probe discriminates rather than blanket-reporting taken. proton.me and zoho.com are refused by Pornhub's own domain allowlist and now return a named error. |
| Do the five omitted modules work elsewhere | **Answered by re-testing from a US exit — see below.** Two of the five were purely geo and work unmodified; the other three fail regardless of location. |

Two bugs surfaced while verifying:

- **github's CSRF pattern was stale and unanchored.** The form now emits `value="…"` *before* `data-csrf="true"`, so `data-csrf="true"\s+value="…"` matched nothing — the fallback could never have fired even with the challenge cleared. The page also carries one token per auto-check endpoint, so a first-match grab risked the `password_validity_checks` token. Now anchored inside the `email_validity_checks` block.
- **pornhub reported disallowed domains as an unexpected API response.** proton.me and zoho.com are refused before any lookup; that is a rule about the provider, not a verdict.

### Re-tested from a US exit

The geo hypothesis was half right. With a New York exit:

| Module | Result |
| --- | --- |
| nextdoor | **Works, unmodified.** DNS resolves (`3.161.213.64`) instead of `0.0.0.0`, confirming the null-route was geo. `invalid_grant` → Registered, `not_found` → Not Registered, both verified over plain httpx. |
| nykaaman | **Works, unmodified.** The Akamai edge serves the site normally; `is_exists` true/false both verified. No transport change needed — httpx is fine. |
| emirates | **Still fails.** The page loads, but the validate-email POST is reset (`HTTP/2 INTERNAL_ERROR`) from the US too, and from inside a real browser same-origin. Akamai bot manager, not geo. |
| annaabi | **Still fails.** Cloudflare managed challenge from the US as well, and it does not clear in a stealth browser. |
| render | **Unchanged, and not location-dependent** — non-enumerable by design. |

So nextdoor and nykaaman were never broken modules; they were unreachable from the author's region. Neither needed a code change to work. The only thing added is a nextdoor diagnostic (`d66e2c6`): its own nameservers answering `0.0.0.0` produced a bare `getaddrinfo failed` that reads like a local network fault, so it now names the cause for anyone scanning from an unsupported region.

### Still not verified

- **pornhub against corporate and self-hosted domains.** Only the major consumer providers were exercised; sub-addressing support elsewhere varies.
- **emirates and annaabi from their own regions.** Both fail from US and the author's region, and both look like bot defence rather than geo, but neither was tested from an EE or AE exit.

### DIFF
--- a/user_scanner/core/impersonate.py
+++ b/user_scanner/core/impersonate.py
@@ -1,3 +1,4 @@
+import asyncio
 import threading
 from typing import Callable, Literal, Optional
 
@@ -55,6 +56,27 @@ def impersonate_request(
     kwargs.setdefault("timeout", _timeout())
     kwargs.setdefault("allow_redirects", False)
     return session.request(method, url, **kwargs)
+
+
+async def impersonate_request_async(
+    url: str,
+    method: Literal["GET", "POST"] = "GET",
+    warmup_url: Optional[str] = None,
+    impersonate: str = DEFAULT_IMPERSONATE,
+    **kwargs,
+) -> cffi.Response:
+    """``impersonate_request`` for the async email modules. curl_cffi's session
+    API is blocking, so it runs in a worker thread; the session cache is shared
+    with the synchronous username modules, so cookies carry over either way.
+    """
+    return await asyncio.to_thread(
+        impersonate_request,
+        url,
+        method,
+        warmup_url=warmup_url,
+        impersonate=impersonate,
+        **kwargs,
+    )
 
 
 def _get_warm_session(
--- a/user_scanner/email_scan/adult/fapfolder.py
+++ b/user_scanner/email_scan/adult/fapfolder.py
@@ -1,4 +1,4 @@
-import httpx
+from user_scanner.core.impersonate import impersonate_request_async
 from user_scanner.core.result import Result
 
 
@@ -32,22 +32,23 @@ async def _check(email: str) -> Result:
     }
 
     try:
-        async with httpx.AsyncClient(timeout=15.0) as client:
-            response = await client.post(url, data=payload, headers=headers)
+        response = await impersonate_request_async(
+            url, "POST", data=payload, headers=headers
+        )
 
-            if response.status_code == 403:
-                return Result.error("403")
+        if response.status_code == 403:
+            return Result.error("403")
 
-            data = response.json()
-            message = data.get("message", "")
+        data = response.json()
+        message = data.get("message", "")
 
-            if "belongs to an existing account" in message:
-                return Result.taken(url=show_url)
+        if "belongs to an existing account" in message:
+            return Result.taken(url=show_url)
 
-            if "password must be at least" in message:
-                return Result.available(url=show_url)
+        if "password must be at least" in message:
+            return Result.available(url=show_url)
 
-            return Result.error(f"Unexpected: {message[:50]}")
+        return Result.error(f"Unexpected: {message[:50]}")
 
     except Exception as e:
         return Result.error(e)
--- a/user_scanner/email_scan/adult/pornhub.py
+++ b/user_scanner/email_scan/adult/pornhub.py
@@ -70,6 +70,14 @@ async def _check(email: str) -> Result:
         if "delivery issues" in error_msg:
             return Result.error(url=SHOW_URL, reason="The email is experiencing email delivery issues")
 
+        # Pornhub refuses to check some providers at all (proton.me, zoho.com),
+        # which is a rule about the domain rather than a verdict on the address.
+        if "is not allowed" in error_msg:
+            return Result.error(
+                url=SHOW_URL,
+                reason=f"Pornhub does not accept registrations from '{domain}'",
+            )
+
         # Sub-addressing is only accepted for mailboxes that actually resolve,
         # so an unusable alias says nothing about the address it derives from.
         if "invalid or cannot be used" in error_msg:
--- a/user_scanner/email_scan/adult/pornhub.py
+++ b/user_scanner/email_scan/adult/pornhub.py
@@ -1,71 +1,87 @@
-import httpx
 import re
+
+from user_scanner.core.impersonate import impersonate_request_async
 from user_scanner.core.result import Result
+
+BASE_URL = "https://www.pornhub.com"
+SHOW_URL = "https://pornhub.com"
+CHECK_API = f"{BASE_URL}/api/v1/user/create_account_check"
+
+# The token is only reachable through the create_account_check URL Pornhub
+# embeds (HTML-escaped) in the page's signup config.
+TOKEN_RE = re.compile(r"create_account_check\?token=([A-Za-z0-9_.\-]+)")
+
+# Pornhub no longer discloses registration for the address as typed: a
+# deliverable address always gets the same "if this email is already
+# registered" placeholder, and that path is what mails the account holder.
+# Probing a sub-addressed alias sidesteps both problems, because Pornhub
+# normalises the "+tag" away and then answers plainly.
+PROBE_TAG = "+phck"
+
+HEADERS = {
+    "x-requested-with": "XMLHttpRequest",
+    "origin": BASE_URL,
+    "referer": BASE_URL + "/",
+    "content-type": "application/x-www-form-urlencoded; charset=UTF-8",
+}
 
 
 async def _check(email: str) -> Result:
-    base_url = "https://www.pornhub.com"
-    show_url = "https://pornhub.com"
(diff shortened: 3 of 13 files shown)


brunolm:
Superseded — split into one PR per fix so each can be reviewed and merged independently.

**Groundwork**
- #528 — `feat(core)`: `impersonate_request_async` (the eight transport fixes below depend on it)

**Independent of the above**
- #529 — wordpress: `email_login_not_allowed` is the exists signal
- #530 — deezer: key the verdict on the email constraint alone
- #531 — nytimes: recognise `show-welcome-back`
- #532 — nextdoor: name the geo null-route

**Depend on #528**
- #533 — kick · #534 — fapfolder · #535 — patreon · #536 — letterboxd
- #537 — classmates · #538 — walmart · #539 — pornhub · #540 — github

Nothing was dropped in the split and no code changed; the branches are rebased onto current `main` (which now includes #525). The five modules discussed here as unreachable — render, annaabi, nykaaman and emirates — remain deliberately unshipped, with one exception: nextdoor and nykaaman were later re-tested from a US exit and turned out to work unmodified, so only nextdoor's diagnostic (#532) survives and nykaaman needs nothing at all.

PR #533, opened 2026-08-07
fix(kick): route the signup check through the impersonating transport

brunolm:
httpx is fingerprint-blocked by Kick's Cloudflare edge, so every address came back as a 403. The impersonating transport clears it and both existing branches resolve unchanged.

```
registered address  -> 422 {"errors":{"email":["The email has already been taken."]}}   Registered
unknown address     -> 204 No Content                                                   Not Registered
```

Both live-tested.

---

### DIFF
--- a/user_scanner/core/impersonate.py
+++ b/user_scanner/core/impersonate.py
@@ -1,3 +1,4 @@
+import asyncio
 import threading
 from typing import Callable, Literal, Optional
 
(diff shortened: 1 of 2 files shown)


kaifcodec:
@brunolm #528 Has been merged.

kaifcodec (approved):
And I tested these ones LGTM.

PR #538, opened 2026-08-07
fix(walmart): clear the 412 session mismatch

brunolm:
httpx is fingerprint-blocked and every address returned `412 Precondition Failed`. Two things beyond the transport swap were needed:

- **The hand-set `User-Agent` and client hints re-tripped the same 412.** They contradicted the impersonated TLS fingerprint, so they are dropped and the transport supplies a self-consistent set.
- **The warm-up cannot use the shared `warmup_url`.** That warms once per session key, but Walmart's login-page URL carries this call's PKCE challenge, so it has to be fetched per check.

```
registered address  -> signInPreference: PASSWORD                Registered
unknown address     -> errors: [{code: EMAIL_ID_INVALID}]        Not Registered
```

Both live-tested. The GraphQL query and verdict logic are unchanged.

---

### DIFF
--- a/user_scanner/core/impersonate.py
+++ b/user_scanner/core/impersonate.py
@@ -1,3 +1,4 @@
+import asyncio
 import threading
 from typing import Callable, Literal, Optional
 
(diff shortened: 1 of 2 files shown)


kaifcodec:
@brunolm I just noticed `walmart` likely returns 412 status code in case of region blocks too.
I think you should change the error message to something more general that covers both the reasons.

kaifcodec (approved):

PR #540, opened 2026-08-07
fix(github): find accounts via the search API when signup is challenged

brunolm:
## tl;dr

- **`/signup` is behind a DataDome interstitial**, so the authenticity_token was unreachable and every address errored.
- **The search API now leads** and confirms a hit outright, with profile metadata.
- **A miss is not proof of availability** — the module says so rather than returning Not Registered.
- **The signup fallback's CSRF pattern was stale and would never have matched**, even with the challenge cleared.

## /signup is behind a DataDome interstitial

Reproduced on `chrome`, `chrome131`, `chrome136`, `firefox135`, `safari184` and `edge101`, with and without the client hints the response asks for. No HTTP client clears it.

## The search API now leads

```
GET https://api.github.com/search/users?q=<email>+in:email
```

A hit is definitive and carries `login`, `id`, `html_url`, `type` and the avatar. It costs one request and uses a **separate rate-limit bucket** from the core API, so it does not compete with `githubgist` for the 60/hr budget.

## A miss is not proof of availability

The API only sees addresses published on a public profile. On a miss the module falls through to the signup probe, and when that is challenged it returns an error naming both reasons instead of a verdict. Behind DataDome the module can therefore confirm Registered but will never return Not Registered — that is a deliberate refusal to guess, not a silent failure.

## The signup fallback's CSRF pattern was stale

GitHub emits `value="…"` **before** the `data-csrf` marker, so `data-csrf="true"\s+value="([^"]+)"` matched nothing — the fallback was dead code regardless of the challenge. The page also carries one token per auto-check endpoint, so a first-match grab risked the `password_validity_checks` token. The search is now anchored inside the `email_validity_checks` block.

Both markers were confirmed against the live endpoint from a browser session that cleared the challenge:

```
unregistered -> 200  "Email is available"
registered   -> 422  "already associated with an account"
```

## Not verified

The available branch has never executed **through the module** — only through a browser. The markers and the corrected pattern are confirmed against live responses, but whether the path runs at all depends on the user's IP.

---

### DIFF
--- a/user_scanner/core/impersonate.py
+++ b/user_scanner/core/impersonate.py
@@ -1,3 +1,4 @@
+import asyncio
 import threading
 from typing import Callable, Literal, Optional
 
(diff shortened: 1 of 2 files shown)


kaifcodec (approved):
@brunolm I also had this approach on my mind.
Even though this is a downgrade but still this is the last option as I also tried to bypass the datadome with stealth-browsers but it is still blocking almost every try!
**MERGING IT!**

kaifcodec:
Maybe we need to keep an eye on it if we can ever find better alternatives! Until then it's good!

PR #548, opened 2026-08-07
fix(donatello): confirm a creator page instead of returning taken on any 200

brunolm:
## tl;dr

- **Reaching the site from a permitted region exposed a false positive**: `/about`, `/login` and `/terms` answer 200 and were reported as **taken handles**.
- **A hit is now confirmed on the creator page's own `og:url`.**
- The Cloudflare regional block is still named rather than surfaced as a bare 403.
- Misses genuinely 404, so the availability branch was already correct.

## The false positive

The module returned `Result.taken()` on any 200. Donatello's own pages answer 200:

```
/about   200   og:url = https://donatello.to        <- site page, was reported TAKEN
/login   200   og:url = https://donatello.to        <- site page, was reported TAKEN
/terms   200   og:url = https://donatello.to        <- site page, was reported TAKEN
```

## A hit is now confirmed on og:url

Only a creator page carries its own handle there:

```
/sternenko  200  og:url = https://donatello.to/sternenko   -> Found
/help       200  og:url = https://donatello.to/help        -> Found   (a real user named "help")
/about      200  og:url = https://donatello.to             -> error
/zzznotarealuser99xzq  404                                 -> Not Found
```

A 200 without it returns an error rather than a verdict.

Live through the module:

```
sternenko             -> Found
about                 -> Error (200 without a creator page, cannot confirm the handle)
zzznotarealuser99xzq  -> Not Found
```

## The regional block

Cloudflare serves "Attention Required" to whole regions; that is a network wall, not a verdict, and is still reported as such. The module also moves to the impersonating transport. Both branches are now live-verified from a region the site serves — the earlier revision of this PR could not exercise them at all.

### DIFF
--- a/user_scanner/user_scan/donation/donatello.py
+++ b/user_scanner/user_scan/donation/donatello.py
@@ -5,14 +5,23 @@ from user_scanner.core.result import Result
 
 CLOUDFLARE_BLOCK_MARKER = "Attention Required! | Cloudflare"
 
+# Only a creator page carries its own handle in og:url; the site's own pages
+# (/about, /login, /terms) answer 200 with the bare domain there instead.
+OG_HANDLE_RE = re.compile(r'<meta property="og:url" content="https://donatello\.to/([^"/?]+)"')
+AUTHOR_RE = re.compile(r'<meta name="author" content="([^"]+)">')
+
 
 def validate_donatello(user):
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):

kaifcodec:
@brunolm Good catch here.

PR #553, opened 2026-08-07
fix(medium): match the served handle and follow subdomain redirects

brunolm:
Medium answers every httpx request through Cloudflare with a 403, so no verdict was reachable. Routed through the impersonating transport — which then exposed three verdict faults:

| Fault | Effect |
| --- | --- |
| Miss inferred from the absent found marker | Any markup change reports every handle free |
| Exact-case match against the requested handle | `QuincyLarson` missed, `quincylarson` found |
| 301 to `{user}.medium.com` unfollowed | Custom-subdomain accounts missed entirely |

A miss is now confirmed on Medium's explicit 404 text; the served handle is compared rather than the requested one, since the meta tag echoes canonical casing; and the subdomain redirect is followed — unknown handles never emit it.

Also extracts name, bio, followers and avatar.

---

Split out of #523; the file here is byte-identical to what was tested there, and `ruff`/`mypy` pass on this branch.

### DIFF
--- a/user_scanner/user_scan/creator/medium.py
+++ b/user_scanner/user_scan/creator/medium.py
@@ -34,13 +34,20 @@ def validate_medium(user: str) -> Result:
         try:
             result = view(user)
         except Exception as e:
-            failures.append(f"{label}: {e}")
+            err_msg = str(e).strip() or type(e).__name__
+            failures.append(f"{label}: {err_msg}")
             continue
         if result:
             return result.update(url=show_url)
 
     if failures:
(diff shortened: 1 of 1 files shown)


kaifcodec:
@brunolm For this module during testing I noticed that when scanning an invented username from here, it correctly reports the username is not taken but if it is a valid username it returns 403, so likely it doesn't let everyone look into the users public data for some regions.

kaifcodec (approved):

PR #557, opened 2026-08-08
fix(annaabi): name the Cloudflare challenge instead of emitting a bare 403

brunolm:
The module reported a bare `Unexpected response status: 403` for every address, which reads like a module bug. It is a Cloudflare managed challenge — `cf-mitigated: challenge` on every request, confirmed from two different countries and across httpx, curl_cffi and urllib.

Two changes: the module moves to the impersonating transport, which may be all it needs where the site is reachable, and the challenge is named when it is not.

## The verdict logic is verified, not assumed

The challenge was cleared by hand in a real browser and both markers still match exactly:

```
test@gmail.com              -> "Sellise emailiga kasutaja on juba registreerinud"   taken
admin@annaabi.ee            -> "See email on vaba"                                  free
info@annaabi.ee             -> "See email on vaba"                                  free
xkqpzrmvn4471zzq@gmail.com  -> "See email on vaba"                                  free
```

So this is a network wall rather than rotted parsing, and the module should work unchanged wherever the site talks to an HTTP client.

## Not verified

It has not run end-to-end from here — the challenge is served to every HTTP client regardless of region, so from my locations it returns the named error rather than a verdict. What is verified is the verdict logic, against real server responses captured once the challenge was cleared.

This mirrors #547, which does the same for the `user_scan` twin.

---

### DIFF
--- a/user_scanner/core/impersonate.py
+++ b/user_scanner/core/impersonate.py
@@ -1,3 +1,4 @@
+import asyncio
 import threading
 from typing import Callable, Literal, Optional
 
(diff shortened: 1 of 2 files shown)


kristoisberg:
I investigated these modules a few days ago when they were consistently failing for me too. Impersonation did not help, but now the modules work for me again all of a sudden. I am in favour of the improved error message, but I'm not sure if the impersonation helps at all.

kaifcodec:
@kristoisberg if this works for you, like if you are able to get a +ve/-ve hits in this test then it's fine, as I am unable to test it properly because this website is blocked in my region.

Anyways about the impersonation I don't think it matters what is used here, because httpx is brittle in these cases as we saw in all the recent fixes we had to migrate from httpx to curl_cffi but we didn't have to do vice versa. So I think using impersonation is just being safe!

kristoisberg (approved):

PR #560, opened 2026-08-09
feature: Tinder username module

kristoisberg:
This PR adds an username module for Tinder. It can only find accounts with a public handle, which is a small subset of all accounts, but this is also the extent to which other tools go.

Accounts for testing can be acquired from yandex.com with the search term `site:tinder.com "log in to like me"` as Google does not index profiles.

### DIFF
--- a/user_scanner/user_scan/social/tinder.py
+++ b/user_scanner/user_scan/social/tinder.py
@@ -1,0 +1,100 @@
+import json
+import re
+
+import httpx
+
+from user_scanner.core.orchestrator import generic_validate
+from user_scanner.core.result import Result
+
+
+def validate_tinder(user: str) -> Result:
+    url = f"https://tinder.com/@{user}"
+
+    def process(response: httpx.Response) -> Result:
+        if response.status_code != 200:
+            return Result.error(f"Unexpected response status: {response.status_code}")
+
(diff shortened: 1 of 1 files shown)


kaifcodec:
> "This PR adds an username module for Tinder. It can only find accounts with a public handle, which is a small subset of all accounts".

@kristoisberg Does it mean most of the profile stays private and they are not accessible in app or web as well? Or just not to be Crawled by bots?

kristoisberg:
> > "This PR adds an username module for Tinder. It can only find accounts with a public handle, which is a small subset of all accounts".
> 
> @kristoisberg Does it mean most of the profile stays private and they are not accessible in app or web as well? Or just not to be Crawled by bots?

They're available both on app and the web, but they haven't claimed a public handle, so they can't be viewed unless they're near you, and don't have a uniquely identifiable username.

kaifcodec (approved):
@kristoisberg The site category doesn't match logically but right now it is the closest we can get without creating a new directory just for only one module.
If you are able to get more of these then we can create a new dedicated, category `dating` for these.
I once had tried but I had given these type of modules low effort as I thought most of the sites likely use extreme bot detection.

Anyways this PR LGTM.

PR #561, opened 2026-08-09
feat: responsive grid layout and alphabetical sorting for list flags

kaifcodec:
- Display user and email module listings (-lu and -le) in a responsive rich Table grid filling terminal size
- Render module items inside panels with inner Table grids for pixel-perfect vertical column alignment
- Sort categories and site modules inside each category 100% alphabetically
- Self-adapt panel height for large categories to eliminate massive vertical gaps and scrolling

### DIFF
--- a/user_scanner/__main__.py
+++ b/user_scanner/__main__.py
@@ -203,15 +203,80 @@ def main():
 
     if args.list_user or args.list_email:
         categories = load_categories(args.list_email, args.no_nsfw)
-        for cat_name, cat_path in categories.items():
-            modules = load_modules(cat_path)
-            print(Fore.MAGENTA + f"\n== {cat_name.upper()} SITES =={Style.RESET_ALL}")
-            for module in modules:
-                name = get_site_name(module)
-                loud = (
-                    f" {R}(loud){X}" if is_loud(name, is_email=args.list_email) else ""
+        sorted_cats = sorted(categories.items(), key=lambda x: x[0].lower())
+        try:
+            import math
+            from rich.console import Console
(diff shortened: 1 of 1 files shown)


kaifcodec:
@brunolm and @kristoisberg try this list layout once and let me know how it looks and whether you think it can be improved or not, right now, I somehow managed to make it look good and level with each column but as the module counts are different in each category it was hard to structure and level them side by side.

kristoisberg:
I tried it out and the responsiveness works great between 2 and 3 columns. I wonder whether a single-column layout would be beneficial or not, e.g. in case the user has multiple side-by-side terminals open, so each one is very narrow. Currently, the two-column layout struggles with rendering loud email modules, but it also fits a lot more information than a single-column layout would.

image

Also, if a category in one column is significantly longer than the category next to it, it creates a large area of empty space under the other one. Did you try out any options where the columns wouldn't need to be at the same level?

kaifcodec:
@kristoisberg Regarding the single-column view, yeah, @VamatoHD mentioned it earlier via Discord. I’m planning to make it fallback to a single column whenever the terminal width is under 70.

As for the empty space issue, I tried a few ways to clean that up, but solving it broke the alphabetical sorting layout. If you have any ideas on how to handle the extra spacing while preserving the alphabetical order, feel free to share or drop a suggestion!

json-hunter07:
@kaifcodec I think keeping alphabetical order and filling up those spaces will not be possible, both of them are not likely possible in the same time.

json-hunter07 (approved):

PR #562, opened 2026-08-09
fix(pypi): stop reporting a package's publisher as the account holder

brunolm:
## tl;dr

- **PyPI reported a package's publisher as the account holder.** `author_email` was emitted as `email` and `author` as `display_name`, so scanning `hugovk` published a defunct company's address and name as that account's own.
- **It now credits every publisher its packages name, under keys that say where they came from** — `author`, `author_email`, `maintainer`, `maintainer_email`.

## Package metadata is not the account holder's

Before. Both values come from `PIL` — first package alphabetically, published by
Secret Labs AB in the 1990s — and the account owner appears nowhere:

```
  [✔] Pypi (hugovk): Found
      ├── display_name: Secret Labs AB (PythonWare)
      ├── email: info@pythonware.com
      ├── packages_count: 24
      └── packages: ['PIL', 'blurb', 'em-keyboard', 'fino', 'flake8-implicit-str-concat']
```

After:

```
  [✔] Pypi (hugovk): Found
      ├── packages_count: 24
      ├── packages: ['PIL', 'blurb', 'em-keyboard', 'fino', 'flake8-implicit-str-concat']
      ├── author: Secret Labs AB (PythonWare), Larry Hastings, Kenneth Reitz, Hugo van Kemenade, Dylan Turner
      ├── author_email: info@pythonware.com, larry@…, me@…, 58230987+…@users.noreply.github.com
      ├── maintainer: Python Core Developers, Hugo van Kemenade
      └── maintainer_email: core-workflow@mail.python.org
```

Three changes produce that:

| Change | Effect |
| --- | --- |
| `email` → `author_email` / `maintainer_email`, `display_name` → `author` / `maintainer` | The key names its source, so no consumer reads it as the account's own address or name |
| Every sampled package is read, not just up to the first hit | Five publishers surface instead of whichever one sorts first; the account owner is now among them |
| Contacts parsed with `email.utils.getaddresses` | `"A <a@x>, B <b@y>"` and `"Thomas Kluyver & contributors <thomas@…>"` no longer mangle |

Cost: `min(5, packages)` requests every time, where a lucky first package used to
cost 1. That is the ceiling the old loop already hit whenever it found nothing.

## Keys changed

| Removed | Added |
| --- | --- |
| `email`, `display_name` | `author`, `author_email`, `maintainer`, `maintainer_email` |

Neither removed key has another reader in the repo — the `display_name` hits in
`core/orchestrator.py` and `core/email_orchestrator.py` are a local variable
holding a category label. Consumers of the exported JSON/CSV/PDF that key off
the old names need updating.

## Testing

| Handle | Case | Result |
| --- | --- | --- |
| `hugovk` | packages with several publishers | Found; 4 addresses, 5 authors, 2 maintainers |
| `<pypi-user-c>` | package with an `author`, no email | Found; `author` only |
| `<pypi-user-b>` | package with no author, no email | Found; `packages_count` + `packages` only |
| `zzznotarealuser99xzq` | nonexistent | Not Found |

Not covered: `_package_info`'s request-failure path, a package crediting two
distinct `maintainer_email`s, and the PDF exporter — the key rename was checked
by `grep`, not by rendering every format.

### DIFF
--- a/user_scanner/user_scan/dev/pypi.py
+++ b/user_scanner/user_scan/dev/pypi.py
@@ -1,20 +1,30 @@
 import re
 import xmlrpc.client
+from email.utils import getaddresses
 from typing import Any
+
 import httpx
 
 from user_scanner.core.helpers import get_random_user_agent
 from user_scanner.core.orchestrator import Result, make_request
+
+# Packages listed on the result, and the ones sampled for release metadata.
+PACKAGE_SAMPLE = 5
+
+# Release metadata names whoever published a package — a co-maintainer, a
+# company, a mailing list — not necessarily the account it hangs off. The keys
+# below keep that provenance, so nothing downstream reads an address here as
+# the account holder's own mailbox or a name here as the account holder's name.
+CONTACT_ROLES = ("author", "maintainer")
 
 
 def validate_pypi(user: str) -> Result:
     """
     Validates a PyPI username and extracts:
 
-    - display_name
-    - email
     - packages_count
     - packages
+    - author / author_email / maintainer / maintainer_email of those packages
     """
 
     if not re.match(r"^(?!_+$)[A-Za-z0-9._-]+$", user):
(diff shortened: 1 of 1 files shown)


kaifcodec:
@brunolm this is ready for review?

brunolm:
I don't think it's ready, I haven't had much time lately. I'll close it for now I guess.

PR #566, opened 2026-08-10
feat(liberapay): extract the OAuth-verified accounts the profile vouches for

brunolm:
A Liberapay profile publishes the accounts its owner proved control of on other platforms, under the site's own wording:

> owns the following accounts on other platforms:

`validate_liberapay` extracted only `name` and discarded that block, even though the module already downloads the page it lives on. It is now emitted as `verified_accounts`.

```
Syncthing  ->  GitHub: https://github.com/syncthing, Twitter: https://x.com/syncthing
Fla        ->  GitHub: …/Flaburgan, Mastodon: https://mastodon.social/@fla,
               Twitter: …/flaburgan, LinuxFr.org: …/users/flaburgan
Changaco   ->  GitHub: …/Changaco, Mastodon: https://diaspodon.fr/@Changaco,
               GitLab.com: …/Changaco, LinuxFr.org: …/users/changaco
```

## Why `verified_accounts` and not a plain link field

Liberapay's `elsewhere` system is OAuth-only — each platform is a `PlatformOAuth2` subclass upstream — so the site is vouching for ownership rather than repeating a URL the owner typed into a bio. That is a materially stronger claim than a profile link, and the key name is what carries the distinction to anything consuming the metadata.

No JSON route carries this. `/<user>/public.json` returns only avatar, display_name, giving, goal, id, kind, npatrons, receiving, statements, summaries and username — the block exists solely in the rendered page, which the module was already fetching. So there is no additional request.

## The parsing is anchored, because the page also lists other people

A Liberapay profile renders patron and donation relationships alongside the owner's own accounts. A document-wide sweep for profile links would harvest strangers and attribute them to the target.

Extraction therefore locates the "owns the following accounts on other platforms" sentence, takes the first `<div class="accounts">` after it, and walks `<div>`/`</div>` tokens with a depth counter so the block is bounded exactly. Anchors are matched only inside that slice. Every failure path — sentence missing, div missing, unbalanced markup — returns empty, so a Liberapay redesign degrades to "no accounts" and can never flip a Found/available verdict or leak a patron.

Mastodon entries are read from the anchor `href` rather than the visible text, since the latter renders as `fla@mastodon.social` and is not a URL.

## Testing

Verified live against `Syncthing`, `Fla` and `Changaco` (all match the rendered pages), an account with no elsewhere block (key absent), and a nonexistent handle (unchanged `available`). `LinuxFr.org` is included deliberately as a platform with no route mapping — it is emitted as a link and simply yields no pivot rather than erroring.

## Not verified

This is HTML scraping, not an API, so it is inherently exposed to a Liberapay redesign. The failure modes above are the mitigation, but the markup itself is untested against any future layout.

### DIFF
--- a/user_scanner/user_scan/donation/liberapay.py
+++ b/user_scanner/user_scan/donation/liberapay.py
@@ -1,4 +1,4 @@
-from user_scanner.core.orchestrator import generic_validate, Result
+from user_scanner.core.orchestrator import generic_validate, make_request, Result
 import html
 import re
 
@@ -37,6 +37,7 @@ def validate_liberapay(user):
     def process(response):
         if response.status_code == 200:
             extra = {}
+            media = {}
             title_match = re.search(r'<title>([^<]+)</title>', response.text)
             if title_match:
                 name = title_match.group(1).split("&#39;s profile")[0].split("'s profile")[0].strip()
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@brunolm I enriched the data extraction of this module and it's now better and perfect.
LGTM!

json-hunter07 (approved):
@kaifcodec Yeah that's perfect.

PR #571, opened 2026-08-12
fix: resolve core engine concurrency hangs

kaifcodec:
This PR addresses several critical issues causing scans to hang or freeze mid-execution:

- **Global Lock Stalls in Impersonate Requests:** Implemented fine-grained `_key_locks` in `impersonate.py` so that warmup requests only block threads using the exact same proxy and impersonation profile, preventing a single slow warmup from freezing the entire engine.
- **Thread Pool Starvation from Dropped Connections:** Wrapped module execution in `asyncio.wait_for` to strictly enforce execution timeouts. Additionally, decoupled the semaphore limit from the underlying thread pool, scaling it dynamically (up to 250 threads by default) so orphaned threads don't starve the concurrency engine.
- **Asynchronous Email Scanning:** Wrapped email modules in `asyncio.create_task` to schedule them immediately on the global event loop, enabling true global concurrency.
- **Real-time Smooth Progress Bar:** Attached asyncio callbacks (`task.add_done_callback`) so the progress bar updates smoothly the millisecond any background task finishes, rather than stuttering between categories.

### DIFF
--- a/tests/test_orchestrator.py
+++ b/tests/test_orchestrator.py
@@ -58,7 +58,7 @@ def test_set_concurrency():
     original_max = orchestrator.MAX_CONCURRENT_REQUESTS
     set_concurrency(10)
     assert orchestrator.MAX_CONCURRENT_REQUESTS == 10
-    assert orchestrator._shared_executor._max_workers == 10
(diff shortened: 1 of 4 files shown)


5dwz5n6d7w-sudo (commented):
Thanks

PR #572, opened 2026-08-12
feature: Andele Mandele username module

kristoisberg:
This PR adds a username module for the Latvian marketplace Andele Mandele (also present in other Baltic countries, but not very popular). The module turned out more complex than I would like, as the site is built somewhat strangely.

- By default, the module fetches enrichments by performing another request against a product the user has listed. This page contains a detailed and standardised JSON object about the user.
- If the user hasn't listed any products, the module falls back to information available on the profile itself. This information is less detailed and often complex to parse (dates in Latvian), which is why they are left in their original form.

### DIFF
--- a/user_scanner/user_scan/shopping/andelemandele.py
+++ b/user_scanner/user_scan/shopping/andelemandele.py
@@ -1,0 +1,160 @@
+import html
+import json
+import re
+from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@kristoisberg Tested the module, LGTM.
**MERGING IT!**

PR #575, opened 2026-08-13
fix: ensure category headers print correctly with --all flag

kaifcodec:
Fixes a UI bug where passing the `--all` (`-a`) flag would completely silence the printing of `== CATEGORY ==` headers in the terminal because the check explicitly skipped printing them. Headers will now print cleanly for visible results even if `--all` is set.


json-hunter07 (approved):
LGTM.

PR #582, opened 2026-08-19
feat: Throne username module

kristoisberg:
This PR adds an username module for throne.com. A few notes:
- The username restrictions are extracted from their client-side bundle.
- Next.js can handle redirects in two different ways, hence the two different availability paths.
- `__NEXT_DATA__` parsing is becoming so common across modules that having Next-specific helpers could reduce boilerplate a lot. I will experiment with this soon.


kristoisberg:
Closing this since the module got merged as part of #583.

PR #585, opened 2026-08-20
feat: CodePen username module

kristoisberg:
CodePen username module, checking both individual users and teams. Performs a separate GraphQL request to fetch information about the user/team.

There is one known limitation to this implementation: CodePen allows individual users and teams to share names, for example:
- https://codepen.io/shopify
- https://codepen.io/team/shopify

In this case, the module prefers the individual user. Is this an issue at all? Do we need to support teams? Should we have separate modules for them? Any other suggestions?

### DIFF
--- a/user_scanner/user_scan/dev/codepen.py
+++ b/user_scanner/user_scan/dev/codepen.py
@@ -1,0 +1,111 @@
+import html
+import re
+from urllib.parse import quote
+
+from user_scanner.core.orchestrator import generic_validate, make_request
+from user_scanner.core.result import Result, Status
+
+
+PROFILE_QUERY = """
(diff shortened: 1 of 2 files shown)


kaifcodec:
If their namespaces are separate, it's definitely better to split this into two modules, `codepen.py` for individual users and `codepenteams.py`.
But if CodePen has some rules like a user who owns the username `shopify` is only able to create a teams account with that same username then this module is fine.

kristoisberg:
I moved teams to a separate module since there are no such restrictions on team names.

`user-scanner -u shopify -m codepen,codepenteams`

kaifcodec (approved):

PR #586, opened 2026-08-20
feat: BugCrowd username module

kristoisberg:
BugCrowd username module. Quite a simple one, there were opportunities to add additional enrichments (i.e. statistics about found bugs), but they would have required additional HTTP requests and don't hold much OSINT value.

### DIFF
--- a/user_scanner/user_scan/dev/bugcrowd.py
+++ b/user_scanner/user_scan/dev/bugcrowd.py
@@ -15,6 +15,9 @@ def validate_bugcrowd(user: str) -> Result:
         if response.status_code == 404 and data.get("error") == "Not Found":
             return Result.available()
 
+        if response.status_code == 404 and not data:
+            return Result.taken(extra={"public": False})
+
         if response.status_code != 200:
             return Result.error(f"Unexpected response status: {response.status_code}")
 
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@kristoisberg the module looks good.

But for some non-existent usernames (like `johndoe`), Bugcrowd's API returns a completely empty JSON dictionary `{}` alongside the 404 status, rather than the expected `{"error": "Not Found"}`. This was causing the validator to throw an unexpected 404 error on those usernames.
I went ahead and pushed a tiny fix to the condition to accept both as valid markers.

Take a look at the commit (9f9459e) by yourself and let me know if this approach was correct.

kristoisberg:
@kaifcodec To me it seems that the empty responses indicate non-public profiles instead. The following page can be used to obtain usernames. 
https://bugcrowd.com/crowdstream?filter_by=accepted,disclosures&page=1

Using a public (clickable) username with the API endpoint results in the full response payload, while a private (non-clickable) username results in an empty response.

@kaifcodec May I force push to remove your commit? I will mark the username as taken instead, with a special marker.

kaifcodec:
My bad, I didn't consider that, yeah go ahead and force push.

kaifcodec (approved):

PR #589, opened 2026-08-20
feat: DevHunt username module

kristoisberg:
This PR adds an username module for https://devhunt.org/. The module uses the site's Supabase API with a public key expiring in 2033.


kaifcodec (approved):
LGTM.

Thank you @kristoisberg .

PR #590, opened 2026-08-21
Add TryHackMe username validator to user_scan/dev

rivalsxninjax1:
Adds a user_scan/dev validator for TryHackMe (tryhackme.com).

Notes on implementation:
- TryHackMe sits behind Vercel's bot-mitigation firewall, which blocks
  plain httpx/curl requests on TLS-fingerprint grounds (429 with
  x-vercel-mitigated: challenge). Used impersonate_validate to route
  through a Chrome-TLS-impersonating session.
- The /p/<username> HTML page is a client-rendered shell identical for
  existing and nonexistent users (data loads client-side via JS), so
  scraping it produces false positives. Instead this hits the public
  JSON API the frontend itself calls:
  GET /api/v2/public-profile?username=<user>
- Nonexistent users return HTTP 404 from this endpoint when
  unauthenticated; existing users return 200 with a rich profile
  payload (level, points, rank, rooms completed, streak, country,
  avatar) — all surfaced via extra/media.

Tested against a known-existing account, a clearly nonexistent one,
and a single-character edge case — all returned correct results:

[✔] Tryhackme (ninjax11): Found
      ├── level: 13
      ├── points: 20594
      ├── rank: 33073
      ├── rooms_completed: 147
      ├── streak: 91
      ├── country: np
      └── avatar: https://cdn-images.tryhackme.com/...
[✘] Tryhackme (some_username_that_almost_certainly_doesnt_exist_xyz123): Not Found

ruff check passes clean.

### DIFF
--- a/user_scanner/user_scan/dev/tryhackme.py
+++ b/user_scanner/user_scan/dev/tryhackme.py
@@ -1,0 +1,86 @@
+import json
+
+from user_scanner.core.impersonate import impersonate_validate
+from user_scanner.core.result import Result
+
+
+def validate_tryhackme(user: str) -> Result:
+    """
+    Checks TryHackMe username availability via the public-profile API.
+
+    TryHackMe sits behind Vercel's bot-mitigation firewall, which blocks
+    plain httpx/curl requests on TLS-fingerprint grounds (observed as an
+    HTTP 429 with an `x-vercel-mitigated: challenge` header, not a real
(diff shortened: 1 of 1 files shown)


kaifcodec (approved):
@rivalsxninjax1 Thank you for the PR, This is is working fine.
Feel free to contribute further.

PR #591, opened 2026-08-21
feat: character.ai username module

kristoisberg:
Username module for [character.ai](https://character.ai/). Uses the site's internal tRPC API as the Next.js frontend is very slow for large profiles, timing out with the default 5 timeout. While the timeouts affected only a very small number profiles, the tRPC approach should be faster for every other profile too.


kaifcodec (approved):
LGTM.
Thank you @kristoisberg .

PR #592, opened 2026-08-21
feat: Beacons username module

kristoisberg:
Beacons.ai username module. There was more enrichment data available on the page, but extracting that would have required more complex React chunk parsing than seemed reasonable.

Potential test cases: `site:beacons.ai`


kaifcodec (approved):
@kristoisberg LGTM.
Thank you!

PR #594, opened 2026-08-22
feat: add Model Context Protocol (MCP) server support

kaifcodec:
- Adds optional mcp dependency group to pyproject.toml
- Registers user-scanner-mcp as a dedicated command line entry point
- Implements modular architecture with schemas.py and handlers.py
- Exposes scan_username and scan_email tools with full parameter support
- Implements cross-scan pivoting configuration via JSON arguments
- Exposes list_available_modules tool for dynamic AI module discovery
- Implements proper stderr logging with verbose output flag support
- Supports AI proxy injection and global concurrency limits natively

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ keywords = ["osint", "email-osint", "username-osint", "python", "cli-tool", "cyb
 [project.optional-dependencies]
 dev = ["pytest", "pytest-cov", "pytest-anyio", "anyio[trio]"]
 pdf = ["reportlab>=4.0.0", "pillow>=10.0.0", "svglib>=1.5.0"]
-mcp = ["mcp>=1.2.0"]
+mcp = ["mcp>=1.2.0,<2"]
 
 [project.urls]
 Homepage = "https://github.com/kaifcodec/user-scanner"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,12 +26,14 @@ keywords = ["osint", "email-osint", "username-osint", "python", "cli-tool", "cyb
 [project.optional-dependencies]
 dev = ["pytest", "pytest-cov", "pytest-anyio", "anyio[trio]"]
 pdf = ["reportlab>=4.0.0", "pillow>=10.0.0", "svglib>=1.5.0"]
+mcp = ["mcp>=1.2.0"]
 
 [project.urls]
 Homepage = "https://github.com/kaifcodec/user-scanner"
 
 [project.scripts]
 user-scanner = "user_scanner.__main__:main"
+user-scanner-mcp = "user_scanner.mcp.server:main"
 
 [tool.flit.module]
 name = "user_scanner"
--- a/user_scanner/mcp/handlers.py
+++ b/user_scanner/mcp/handlers.py
@@ -1,49 +1,61 @@
 import asyncio
+import contextlib
 import json
 import logging
+import sys
 import mcp.types as types
 from typing import Any
 
-from user_scanner.core import engine
 from user_scanner.core.formatter import into_json
-from user_scanner.core.helpers import find_module, load_categories, load_modules, get_site_name, ScanConfig, set_proxy_manager, set_global_timeout
+from user_scanner.core.helpers import (
+    find_module,
+    load_categories,
+    load_modules,
+    get_site_name,
+    ScanConfig,
+    set_proxy_manager,
+    set_global_timeout,
+)
 from user_scanner.core.cross_scan import run_cross_scan, CrossScanConfig
-from user_scanner.core.orchestrator import set_concurrency as set_user_concurrency
-from user_scanner.core.email_orchestrator import set_concurrency as set_email_concurrency
+from user_scanner.core.orchestrator import (
+    run_user_full,
+    run_user_category,
+    run_user_module,
+    set_concurrency as set_user_concurrency,
+)
+from user_scanner.core.email_orchestrator import (
+    run_email_full_batch,
+    run_email_category_batch,
+    run_email_module_batch,
+    set_concurrency as set_email_concurrency,
+)
+from user_scanner.core.result import Status
 
 logger = logging.getLogger("user-scanner-mcp")
+
+# Default concurrency values (must match orchestrator/email_orchestrator defaults)
+_DEFAULT_USER_CONCURRENCY = 60
+_DEFAULT_EMAIL_CONCURRENCY = 25
+
+# Serialise scans so concurrent tool calls don't stomp global state
+_scan_lock = asyncio.Lock()
+
 
 async def execute_scan(arguments: dict, is_email: bool) -> list[types.TextContent]:
     target_key = "email" if is_email else "username"
     target = arguments.get(target_key)
-    
+
     if not target:
         raise ValueError(f"Missing '{target_key}' argument")
-        
+
     category = arguments.get("category")
     module_name = arguments.get("module")
     allow_loud = arguments.get("allow_loud", False)
     no_nsfw = arguments.get("no_nsfw", False)
     proxies = arguments.get("proxies")
-    
-    if proxies:
-        logger.info(f"Setting up proxy manager with {len(proxies)} proxies.")
-        set_proxy_manager(proxies=proxies)
-    else:
-        # Clear it just in case this is a subsequent request
-        set_proxy_manager(proxies=None)
-        
     timeout = arguments.get("timeout")
-    if timeout is not None:
-        set_global_timeout(float(timeout))
-    else:
-        set_global_timeout(None)  # type: ignore
-        
     concurrency = arguments.get("concurrency")
-    if concurrency is not None:
-        set_user_concurrency(int(concurrency))
-        set_email_concurrency(int(concurrency))
-    
+
     cross_scan = arguments.get("cross_scan", False)
     cross_links = arguments.get("cross_links", "all")
     cross_emails = arguments.get("cross_emails", "verified")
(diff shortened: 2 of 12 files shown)


chrisbrowns201392-create (commented):
Thank

pyproject.toml
```diff
@@ -26,12 +26,14 @@ keywords = ["osint", "email-osint", "username-osint", "python", "cli-tool", "cyb
 [project.optional-dependencies]
 dev = ["pytest", "pytest-cov", "pytest-anyio", "anyio[trio]"]
 pdf = ["reportlab>=4.0.0", "pillow>=10.0.0", "svglib>=1.5.0"]
+mcp = ["mcp>=1.2.0"]
```
brunolm: **Blocker (1):** `mcp = ["mcp>=1.2.0"]` has no upper bound, and `mcp` 2.0.0 removed the decorator API this server is built on, so `pip install "user-scanner[mcp]"` today produces a server that crashes on import. `Server` in 2.0.0 registers handlers through the constructor (`on_list_tools=`, `on_call_tool=`) and has no `list_tools` / `call_tool` methods at all, so `user_scanner/mcp/server.py:27` raises `AttributeError` before `main()` ever runs; pin `mcp>=1.2.0,<2` (matching the bounded style of `httpx[http2]>=0.27,<0.29` in the same file), or port to the 2.0 API — which also means returning `types.CallToolResult` instead of `list[TextContent]`.

Verified against the installed 2.0.0 source: the only occurrences of those names are the `on_list_tools` / `on_call_tool` constructor params and the `("tools/list", …, on_list_tools)` dispatch table.

The `# type: ignore` on `server.py:27` and `server.py:32` is what hides this from CI: `lint-pr.yml` installs `mcp` unpinned, so mypy sees the same 2.0.0 and would have flagged the missing attribute.

**tl;dr (problem):** unpinned `mcp` resolves to 2.0.0, which deleted `@app.list_tools()` / `@app.call_tool()`.
**tl;dr (fix):** pin `mcp>=1.2.0,<2` and drop the `# type: ignore` that hid it.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+    
+    # 1. Run Initial Scan
+    results = []
+    if module_name:
+        modules = find_module(module_name, is_email=is_email, no_nsfw=no_nsfw)
+        if not modules:
+            return [types.TextContent(type="text", text=f"Error: Module '{module_name}' not found.")]
+        # For simplicity in MCP, we just use engine.check which ignores loud checks if directly targeted
+        # But to be safe, we just run the first module
+        result = await engine.check(modules[0], target)
```
brunolm: **Blocker (2):** `allow_loud` is read at line 25 and put in a `ScanConfig` that the initial scan never consults, so every scan runs the loud modules regardless. `engine.check` / `check_category` / `check_all` take no `ScanConfig` and never call `is_loud`; the loud gate lives in the orchestrators (`email_orchestrator.py:89`, `orchestrator._async_worker`) that this handler bypasses. The 28 entries in `LOUD_MODULES["email"]` are exactly the ones that mail the target — `loud_prompt.check_loud_module_permission` refuses to run them non-interactively for that reason — so an agent calling `scan_email(email="victim@…")` with the documented default `allow_loud: false` sends the target a password-reset/verification mail. Route the scan through `run_email_full_batch` / `run_user_full` (as `__main__.py:597` does) instead of `engine.*`.

The comment on line 67 — "engine.check which ignores loud checks if directly targeted" — acknowledges the bypass, but the CLI does not ignore it there either: it prompts, and skips outright when stdin is not a tty.

**tl;dr (problem):** `allow_loud` is accepted and discarded; loud modules always run and mail the target.
**tl;dr (fix):** call `run_user_full` / `run_email_full_batch` with the `ScanConfig` instead of `engine.check*`.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+        cross_configs = CrossScanConfig(
+            links=cross_links,
+            emails=cross_emails,
+            sweep=cross_sweep,
+            depth=cross_depth,
+            modules=(module_name,) if module_name else (),
+            categories=(category,) if category else (),
+        )
+        # run_cross_scan modifies/returns a list of new results
+        cross_results = await asyncio.to_thread(run_cross_scan, results, config, cross_configs)
```
brunolm: **Blocker (3):** `run_cross_scan` writes to stdout, which under stdio transport *is* the JSON-RPC channel, so any call with `cross_scan: true` injects non-JSON bytes into the framing and breaks the session. `cross_scan.run_cross_scan` opens with `print(f"\n{Fore.MAGENTA}== CROSS-SCAN ==…")` and prints per-round headers; the orchestrators it calls print `== <CATEGORY> SITES ==` banners and `Result.show()` → `print(self.get_console_output(configs))` for every TAKEN/SKIPPED hit. The stderr logging set up in `server.py` covers the module's own `logger` calls only, not the engine's `print`s. Wrap the scan in `contextlib.redirect_stdout(sys.stderr)` — `stdio_server` captures `sys.stdout.buffer` once on entry, so rebinding `sys.stdout` afterwards does not disturb the transport.

`rich.Progress` is transient and detects a non-terminal, so the progress bars are not the problem; the plain `print()` calls are. Note the redirect is process-global, so it has to cover overlapping calls too — see finding 5.

**tl;dr (problem):** cross-scan `print()`s into the JSON-RPC stdout stream and corrupts it.
**tl;dr (fix):** run the scan under `contextlib.redirect_stdout(sys.stderr)`.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+        # But to be safe, we just run the first module
+        result = await engine.check(modules[0], target)
+        results = [result]
+    elif category:
+        try:
+            results = await engine.check_category(category, target, is_email=is_email)
+        except ValueError as e:
+            return [types.TextContent(type="text", text=json.dumps({"error": str(e)}))]
+    else:
+        results = await engine.check_all(target, is_email=is_email)
```
brunolm: **Major (4):** `no_nsfw` is discarded for category and full scans, same root cause as finding 2: `engine.check_all` calls `load_categories(is_email=is_email)` and `check_category` does the same, neither forwarding `no_nsfw`, so the `adult` category is always scanned. The CLI gets this right via `load_categories(is_email, args.no_nsfw)` at `__main__.py:535` and `email_orchestrator.py:210`. Only the `module` branch on line 64 honours the flag. Fixed by the same switch to the orchestrators.

**tl;dr (problem):** `no_nsfw: true` still scans the `adult` category on category/all scans.
**tl;dr (fix):** same as finding 2 — go through the orchestrators, which pass `no_nsfw` to `load_categories`.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+    if not target:
+        raise ValueError(f"Missing '{target_key}' argument")
+        
+    category = arguments.get("category")
+    module_name = arguments.get("module")
+    allow_loud = arguments.get("allow_loud", False)
+    no_nsfw = arguments.get("no_nsfw", False)
+    proxies = arguments.get("proxies")
+    
+    if proxies:
```
brunolm: **Major (5):** every request mutates process-global state (`set_proxy_manager`, `set_global_timeout`, `set_concurrency`) while the SDK dispatches tool calls concurrently, so overlapping scans corrupt each other's settings. `Server.run` in mcp 1.x does `tg.start_soon(self._handle_message, …)` per incoming message, so a second `scan_username` starting mid-flight through the first will call `set_proxy_manager(proxies=None)` and `set_global_timeout(None)` on lines 34/40 and strip the running scan of its proxies and timeout. Either serialise `execute_scan` behind an `asyncio.Lock`, or thread the settings through `ScanConfig` instead of globals.

**tl;dr (problem):** concurrent tool calls stomp each other's proxy/timeout/concurrency globals.
**tl;dr (fix):** guard `execute_scan` with an `asyncio.Lock`, or pass settings via `ScanConfig`.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+        # Clear it just in case this is a subsequent request
+        set_proxy_manager(proxies=None)
+        
+    timeout = arguments.get("timeout")
+    if timeout is not None:
+        set_global_timeout(float(timeout))
+    else:
+        set_global_timeout(None)  # type: ignore
+        
+    concurrency = arguments.get("concurrency")
```
brunolm, line 57: **Major (6):** `concurrency` is the one setting with no reset branch, so it leaks across requests in a way `proxies` and `timeout` deliberately do not. Lines 29–34 and 36–40 both restore the default when the argument is absent; lines 42–45 only ever set it, so one call passing `concurrency: 1` throttles every later scan in that server process to a single request at a time, with no way for the client to undo it. Add the symmetric `else` restoring the module defaults (60 for user, 25 for email).

**tl;dr (problem):** `concurrency` is never restored, so one low value throttles the rest of the session.
**tl;dr (fix):** add the `else` branch resetting to the module defaults, like `proxies` and `timeout` do.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+
+    logger.info(f"Executing {'email' if is_email else 'username'} scan for: {target}")
+    
+    # Construct base config
+    config = ScanConfig(allow_loud=allow_loud, no_nsfw=no_nsfw, timeout=timeout)
+    
+    # 1. Run Initial Scan
+    results = []
+    if module_name:
+        modules = find_module(module_name, is_email=is_email, no_nsfw=no_nsfw)
```
brunolm: **Major (7):** the module names `list_available_modules` advertises cannot be fed back into the `module` parameter for the 31 modules whose file name contains an underscore. `get_site_name` renders `made_porn.py` as `Made.porn`, but `find_module` compares against the raw stem `made_porn`, so `module: "made.porn"` returns "not found"; the CLI works around this with `find_module(m.replace(".", "_"), …)` at `__main__.py:518`, which this handler omits. Since the schema explicitly tells the agent to "use list_available_modules to see options", the discovery tool hands out inputs its own scan tool rejects. Apply the same `.replace(".", "_")` normalisation.

Affected stems include `made_porn`, `chess_com`, `bio_link`, `nebula_tv`, `d3_ru` (31 total across both scan types).

**tl;dr (problem):** `list_available_modules` returns `Made.porn`, but `module: "made.porn"` never matches.
**tl;dr (fix):** normalise with `module_name.replace(".", "_")` before `find_module`, as `__main__.py:518` does.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+            modules=(module_name,) if module_name else (),
+            categories=(category,) if category else (),
+        )
+        # run_cross_scan modifies/returns a list of new results
+        cross_results = await asyncio.to_thread(run_cross_scan, results, config, cross_configs)
+        results.extend(cross_results)
+        
+    # 3. Filter and Format
+    # Filter out empty results to save LLM context window limits
+    found = [r for r in results if r.is_found()]
```
brunolm: **Major (8):** filtering to `r.is_found()` discards every error and skip with no counter, so the agent cannot distinguish "the target has no account there" from "that module was blocked, rate-limited or timed out". `is_found()` is `status == Status.TAKEN` only; a `Result.error` from a blocked site is indistinguishable in the output from a clean miss, which is the false-negative failure mode `AGENTS.md` forbids at module level, reintroduced at the transport layer — and an LLM will confidently report a clean sheet. Return an envelope instead: the found results plus per-status counts and the erroring site names.

**tl;dr (problem):** errors and skips are dropped silently, so blocked sites read as "not found".
**tl;dr (fix):** wrap the payload with per-status counts and the list of sites that errored.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+        
+    timeout = arguments.get("timeout")
+    if timeout is not None:
+        set_global_timeout(float(timeout))
+    else:
+        set_global_timeout(None)  # type: ignore
+        
+    concurrency = arguments.get("concurrency")
+    if concurrency is not None:
+        set_user_concurrency(int(concurrency))
```
brunolm: **Minor (9):** calling `orchestrator.set_concurrency` per request leaks threads, because it rebinds `_shared_executor` to a fresh `ThreadPoolExecutor` without shutting the old one down. Worker threads already spawned by the discarded pool block on their queue and are only released at interpreter exit, so in a long-lived server each `concurrency` argument strands another pool's threads — harmless for the CLI, which calls this once per process. Shut down the previous executor before replacing it, or size the semaphore per run instead of swapping the executor.

Pre-existing in `orchestrator.py`, but this PR is the first caller that can invoke it repeatedly.

**tl;dr (problem):** each per-request `set_concurrency` strands the previous thread pool's threads.
**tl;dr (fix):** `shutdown(wait=False)` the old executor before rebinding, or stop swapping it per request.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+        set_proxy_manager(proxies=proxies)
+    else:
+        # Clear it just in case this is a subsequent request
+        set_proxy_manager(proxies=None)
+        
+    timeout = arguments.get("timeout")
+    if timeout is not None:
+        set_global_timeout(float(timeout))
+    else:
+        set_global_timeout(None)  # type: ignore
```
brunolm: **Minor (10):** `set_global_timeout(None)  # type: ignore` silences a real signature mismatch rather than fixing it; `helpers.set_global_timeout` is annotated `timeout: float` while every consumer already handles `None` (`get_global_timeout() or 15.0`, `if global_timeout is not None`). Widen the parameter to `Optional[float]` in `helpers.py` and drop the suppression.

**tl;dr (problem):** `# type: ignore` papers over a wrong annotation on `set_global_timeout`.
**tl;dr (fix):** annotate the parameter `Optional[float]` and remove the ignore.

user_scanner/mcp/schemas.py
```diff
@@ -0,0 +1,107 @@
...
+            description="Lists all available categories and modules (sites) supported by the scanner. Use this to discover valid inputs for the 'category' and 'module' parameters.",
+            inputSchema={
+                "type": "object",
+                "properties": {
+                    "is_email": {
+                        "type": "boolean",
+                        "description": "If true, lists email scan modules. If false, lists username scan modules. Default: False"
+                    }
+                },
+                "required": ["is_email"]
```
brunolm: **Minor (11):** `list_available_modules` declares `is_email` in `required` while its own description says "Default: False", so a client following the description gets a schema-validation failure. Either drop it from `required` (the handler already defaults it at `handlers.py:101`) or delete the "Default" sentence. The same tool also ignores `no_nsfw`, so it advertises the adult modules even to a client that will scan with `no_nsfw: true`.

**tl;dr (problem):** `is_email` is required but documented as optional with a default.
**tl;dr (fix):** remove it from `required`; optionally accept `no_nsfw` for the listing too.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+from user_scanner.core import engine
+from user_scanner.core.formatter import into_json
+from user_scanner.core.helpers import find_module, load_categories, load_modules, get_site_name, ScanConfig, set_proxy_manager, set_global_timeout
+from user_scanner.core.cross_scan import run_cross_scan, CrossScanConfig
+from user_scanner.core.orchestrator import set_concurrency as set_user_concurrency
+from user_scanner.core.email_orchestrator import set_concurrency as set_email_concurrency
+
+logger = logging.getLogger("user-scanner-mcp")
+
+async def execute_scan(arguments: dict, is_email: bool) -> list[types.TextContent]:
```
brunolm, line 44: **Minor (12):** no tests, while every other core surface has one (`tests/test_engine.py`, `test_orchestrator.py`, `test_cross_scan.py`, …). The branches worth covering are pure and need no network: missing target raises, `category` + `module` together raises, unknown tool name raises, and the argument-to-`ScanConfig` mapping — the last of which would have caught findings 2 and 4.

**tl;dr (problem):** the new MCP layer ships with no tests, unlike every other core module.
**tl;dr (fix):** add `tests/test_mcp_handlers.py` covering argument validation and the `ScanConfig` mapping.

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
...
+    target_key = "email" if is_email else "username"
+    target = arguments.get(target_key)
+
```
brunolm: **Nit (13):** trailing whitespace on ~14 lines across `handlers.py` (19, 22, 28, 35, 41, 46, 57, 60, 93, 103, 105, 110) and `schemas.py`. Ruff's default `select` does not include `W`, so CI stays green; still worth stripping.

**tl;dr (problem):** trailing whitespace on ~14 added lines.
**tl;dr (fix):** strip it (`ruff check --select W291,W293 --fix`).

user_scanner/mcp/handlers.py
```diff
@@ -0,0 +1,126 @@
+import asyncio
+import json
+import logging
+import mcp.types as types
+from typing import Any
+
+from user_scanner.core import engine
+from user_scanner.core.formatter import into_json
+from user_scanner.core.helpers import find_module, load_categories, load_modules, get_site_name, ScanConfig, set_proxy_manager, set_global_timeout
```
brunolm: **Nit (14):** the 140-char single-line `from user_scanner.core.helpers import …` breaks with the parenthesized multi-line import style used everywhere else in the codebase (`cross_scan.py:27`, `orchestrator.py:13`, `email_orchestrator.py:9`).

**tl;dr (problem):** one-line 140-char import, against the repo's multi-line import style.
**tl;dr (fix):** wrap it in parentheses like the core modules do.

brunolm (changes requested):
Reviewed the full diff plus the core it calls into. 3 blockers — the server does not start on a fresh install, it ignores `allow_loud`, and cross-scan corrupts the JSON-RPC stream. The rest is per-request global state leaking between calls, plus polish.

**tl;dr:**
1. `pyproject.toml:29` — unpinned `mcp` resolves to 2.0.0, which deleted `@app.list_tools()` / `@app.call_tool()`.
2. `user_scanner/mcp/handlers.py:69` — `allow_loud` is accepted and discarded; loud modules always run and mail the target.
3. `user_scanner/mcp/handlers.py:91` — cross-scan `print()`s into the JSON-RPC stdout stream and corrupts it.
4. `user_scanner/mcp/handlers.py:77` — `no_nsfw: true` still scans the `adult` category on category/all scans.
5. `user_scanner/mcp/handlers.py:29` — concurrent tool calls stomp each other's proxy/timeout/concurrency globals.
6. `user_scanner/mcp/handlers.py:42` — `concurrency` is never restored, so one low value throttles the rest of the session.
7. `user_scanner/mcp/handlers.py:64` — `list_available_modules` returns `Made.porn`, but `module: "made.porn"` never matches.
8. `user_scanner/mcp/handlers.py:96` — errors and skips are dropped silently, so blocked sites read as "not found".
9. `user_scanner/mcp/handlers.py:44` — each per-request `set_concurrency` strands the previous thread pool's threads.
10. `user_scanner/mcp/handlers.py:40` — `# type: ignore` papers over a wrong annotation on `set_global_timeout`.
11. `user_scanner/mcp/schemas.py:104` — `is_email` is required but documented as optional with a default.
12. `user_scanner/mcp/handlers.py:16` — the new MCP layer ships with no tests, unlike every other core module.
13. `user_scanner/mcp/handlers.py:19` — trailing whitespace on ~14 added lines.
14. `user_scanner/mcp/handlers.py:9` — one-line 140-char import, against the repo's multi-line import style.

kaifcodec:
@brunolm If you'd like, feel free to go ahead and push the requested fixes in a new commit to this PR. Otherwise, no worries, I can take care of them tomorrow.

kaifcodec:
@brunolm I merged this PR early as the raised Review are mostly resolved.
You can always write it here if you find any issues on this.

user_scanner/mcp/server.py
```diff
@@ -0,0 +1,70 @@
...
+except ImportError:
+    print("Error: The 'mcp' dependency is not installed.", file=sys.stderr)
+    print("Please install it using: pip install user-scanner[mcp]", file=sys.stderr)
+    sys.exit(1)
+
+from user_scanner.mcp.schemas import get_tool_list
+from user_scanner.mcp.handlers import call_tool
+
+# Configure logging to write to stderr so it doesn't corrupt the JSON-RPC stdout stream
+logging.basicConfig(
```
chrisbrowns201392-create, line 18: Gliliana651@gmail.com

PR #597, opened 2026-08-22
feat: Ko-fi username module

kristoisberg:
This PR adds a new username module for Ko-fi. It also fixes Ko-fi links incorrectly pivoting to the Buymeacoffee module.


kaifcodec (approved):
LGTM.

PR #601, opened 2026-08-22
fix(pdf): isolate batch scans into separate pdf reports

### LINKED ISSUE #596: Reports
When I generate a PDF using multiple usernames:

user-scanner -uf usernames.txt -f pdf -o report.pdf

it doesn’t generate a separate page for each user; instead, it mixes everything together. It would be good if it generated one page per user, containing only that user’s own findings.

kaifcodec:
Fixes #596.

This PR restructures the batch processing and cross-scan execution loops in `__main__.py`. Instead of accumulating all batch targets into a single list and attempting to render them into a single PDF (which caused empty pages and data collision), the engine now natively processes each seed target in isolation. 

**Key Changes:**
- Loops over `targets_found` and fully executes both standard modules and cross-scans per target.
- Tags all results with an internal `seed_target` marker.
- The PDF exporter now natively reads this marker and saves a separate PDF file (e.g. `johndoe_report.pdf`, `kaifcodec_report.pdf`) for every seed target in the batch.
- Cross-scanned pivots remain grouped perfectly underneath the parent seed target that spawned them.

### DIFF
--- a/user_scanner/__main__.py
+++ b/user_scanner/__main__.py
@@ -1,7 +1,7 @@
 import argparse
 import json
-import os
 import sys
+import os
 import time
 
 from colorama import Fore, Style
@@ -545,6 +545,7 @@ def main():
                 validated_categories.append(cat_path)
 
     for i, target in enumerate(targets):
+        target_results = []
         if i != 0 and args.delay:
             time.sleep(args.delay)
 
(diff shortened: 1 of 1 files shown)


json-hunter07 (approved):
LGTM.

PR #603, opened 2026-08-24
feat: guns.lol username module

kristoisberg:
Despite the name, another link-in-bio type site. The site has a very low rate limit before bot challenges kick, but it shouldn't affect individual checks too much, only batch jobs.


kaifcodec (approved):
LGTM, @kristoisberg .

PR #608, opened 2026-08-27
fix(email): update Polarsteps API to v10.6.0

kaifcodec:
Updates the polarsteps `email_scan` module to use their new API endpoint (`/api/users/validation/unique`).

Changes:
- Updates `User-Agent` and custom polarsteps headers to match app version `10.6.0`.
- Dynamically generates a random `polarsteps-device-id` on each request using `secrets.token_hex(8)` to avoid device bans.
- Migrates the parsing logic to handle the new JSON response format (`{"is_unique": true/false}`) rather than the old raw text format.

### DIFF
--- a/user_scanner/email_scan/travel/polarsteps.py
+++ b/user_scanner/email_scan/travel/polarsteps.py
@@ -1,25 +1,30 @@
 import httpx
+import secrets
 from user_scanner.core.result import Result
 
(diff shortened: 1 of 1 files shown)


json-hunter07 (approved):
LGTM. The issue is fixed!

PR #618, opened 2026-08-30
fix: OkCupid email module false positives

kristoisberg:
The OkCupid email module currently produces false positives for all emails under certain domains (I'm currently only aware of mail.ru), this PR fixes that issue. Since the response does not distinguish between taken and otherwise invalid/blacklisted emails, I applied a fix similar to https://github.com/kaifcodec/user-scanner/pull/473, generating a random email address under the same domain, retrying the request, and determining the failure reason based on that.


kaifcodec (approved):
@kristoisberg LGTM.

PR #619, opened 2026-08-30
fix(email_scan): retire render — the signup oracle no longer exists

### LINKED ISSUE #612: render: every address returns {"email":"invalid"} — both verdict branches are now unreachable
`email_scan/hosting/render.py` returns `Render Error: {"email":"invalid"}` for every address, so the module can no longer produce either of its two verdicts.

Reproduced on v1.5.1 from two different networks (residential and datacentre) with three unrelated addresses:

```
lgctrl9f3a2b7c4d5e6f@gmail.com   HTTP 200   '{"email":"invalid"}'
john.smith.4471@gmail.com        HTTP 200   '{"email":"invalid"}'
a1b2c3@outlook.com               HTTP 200   '{"email":"invalid"}'
```

It is not address-shaped and it is not an IP block — the response is identical for well-formed addresses at different providers, and HTTP 200 throughout.

### Cause

The module posts the `signUp` GraphQL mutation and classifies on the error string:

```python
if '"email":"exists"' in msg:
    return Result.taken(url=show_url)
elif '"hcaptcha_token":"invalid"' in msg:
else:
    return Result.error(f"Render Error: {msg}")
```

Raw response for all three addresses:

```json
{"errors":[{"message":"{\"email\":\"invalid\"}","path":["signUp"]}],"data":{"signUp":null}}
```

The oracle depended on Render validating the email *after* the captcha, so a valid-but-unregistered address failed at `hcaptcha_token` and a registered one failed at `email:exists`. It now returns `email:invalid` first, before either branch is reachable — so neither verdict can occur any more and every address falls to the `else`.

I could not find a payload that restores the old ordering; `email` appears to be rejected up front now regardless of what else is sent. If that is right, the signal this module relied on is gone rather than moved, and the honest outcome may be removing it (`abandoned/`) rather than repairing it.

### Suggested handling either way

Whatever the fix, `{"email":"invalid"}` is now a *known* response rather than an unexpected one, and it currently renders as a generic `Error`. Even if the module is retired, mapping this specific message to a clear reason — "Render rejects the address before the existence check; this oracle no longer works" — would stop it reading like a transient failure that might succeed on a retry.

Found while integrating the email side into an OSINT bench, where the distinction between "no answer" and "no account" matters a great deal. Retrying an endpoint that can never answer is wasted budget, so knowing this one is structurally dead rather than flaky is genuinely useful.

ARAVIND281:
Closes #612

The diagnosis in the issue is right, and the conclusion it points at is that the module cannot be repaired. I probed it before accepting that, and it holds.

## The oracle depended on ordering that no longer exists

`render.py` classified on *which* validation error `signUp` returned — `"email":"exists"` meant taken, `"hcaptcha_token":"invalid"` meant valid-but-unregistered. That only worked because Render validated the email **before** the captcha, so an unregistered address fell through to the captcha complaint. It now answers `{"email":"invalid"}` first, and neither branch is reachable.

## Probed live

```
valid unregistered      HTTP 200   {"email":"invalid"}
likely-registered       HTTP 200   {"email":"invalid"}
syntactically invalid   HTTP 200   {"email":"invalid"}
```

The third row is the one that settles it: a **malformed** address returns the same body as a well-formed one, so the response is not a verdict about the address at all.

## Ruled out, so this is not transport or payload

| Hypothesis | Result |
|---|---|
| Payload is missing a captcha field | Adding `hcaptchaToken` changes nothing |
| Anti-bot TLS fingerprinting | `curl_cffi` `impersonate="chrome"`, register page warmed first so cookies carry — byte-identical bodies for all three inputs |
| Mutation renamed or removed | Still exists; a bare `signUp` fails GraphQL **validation**, not "unknown field" |
| Schema changed shape | Introspection is disabled (500 on `__type`) |

## Why it is unrecoverable

Loading `dashboard.render.com/register` in a real browser shows the page pulling `newassets.hcaptcha.com`. The live flow obtains a **solved hCaptcha token** before `signUp` is called, and the API answers anything without one with a single generic error.

So there is no remaining signal separating a registered address from an unregistered one, and recovering one would mean solving a captcha — which a scan module cannot do, and which this project would not want it to.

## What this PR does

Moves the module to `abandoned/email_scan/hosting/render.py` per AGENTS.md — same scan type, same category, same file name, **contents unchanged** — rather than deleting it, so it can be revived if Render ever restores an observable difference. Git records it as a 100% rename.

`bunny` and `neocities` in the same category are untouched.

## Checks

```
ruff check .        All checks passed!
mypy user_scanner   Success: no issues found in 528 source files   (was 529)
pytest              380 passed, 3 skipped
```

`user-scanner -e ... -m render` now reports `Email module 'render' not found` instead of erroring on every address.

## One thing for you to decide

If you would rather keep it listed and have it return a clear `Result.error` explaining the site is captcha-gated — so users who look for Render get an explanation rather than "module not found" — say so and I will switch it. AGENTS.md reads to me as retirement being the intended path for "a module can no longer work", so that is what I did, but it is your call.


kaifcodec (approved):
Makes sense.
MERGING IT.

PR #620, opened 2026-08-31
fix(email_scan): add guest UUID to Lovescape signup endpoint

### LINKED ISSUE #613: lovescape: signup endpoint now requires a guest uuid — every address returns an error
`email_scan/adult/lovescape.py` returns `Unexpected: Guest uuid is required` for every address. The signup endpoint has gained a precondition the module does not satisfy.

Reproduced on v1.5.1 from two different networks with several addresses — it is a precondition failure rather than anything address-dependent.

```
user-scanner -e john.smith.4471@gmail.com -m lovescape --all
[!] Lovescape: Error (Unexpected: Guest uuid is required)
```

### Cause

Replaying the module's own POST to `https://lovescape.com/api/front/auth/signup` with its own headers and payload:

```
HTTP 400
{"error":"Guest uuid is required","data":[]}
```

The request is rejected before the email is evaluated, so the module never reaches a taken/available signal. The API now expects a guest identifier that a browser would have obtained earlier in the flow — most likely a cookie or a value from a prior front-end call — and the module posts to `/signup` cold.

The payload already carries `"fingerprint": ""` and an `ampl` block with `"device_id": ""` and a hardcoded `session_id`, which suggests the site was already moving in this direction when the module was written.

### What would fix it

Establishing whatever issues the guest uuid before the signup POST — presumably a GET to the signup page, or a call to a `/api/front/...` guest or session endpoint, carried on the same client so the cookie persists. I did not chase which specific call mints it, since that means probing an adult site's auth flow more aggressively than I was willing to do for a bug report; someone with the front-end open in devtools will find it in one page load.

If the precondition turns out to need a real browser session, this may belong in `abandoned/` alongside the other modules whose signal has gone rather than moved.

### Small thing regardless of the fix

The reason surfaces as `Unexpected: Guest uuid is required`, which reads as an unknown condition. It is now the *only* thing this module returns, so labelling it as a known unsupported-precondition state would distinguish it from a genuinely unexpected response — useful for anyone deciding whether a retry is worth attempting.

Found while integrating the email side into an OSINT bench.

Azaucifer:
- Lovescape API now requires guest UUID in signup requests
- Add _get_guest_uuid() to fetch UUID from /api/front/config
- Include guestUuid in the signup payload
- Fixes #613

### DIFF
--- a/user_scanner/email_scan/adult/lovescape.py
+++ b/user_scanner/email_scan/adult/lovescape.py
@@ -1,7 +1,7 @@
 import httpx
-import json
 import re
 from user_scanner.core.result import Result
+
 
 async def _get_guest_uuid(client: httpx.AsyncClient) -> str:
     """Get a guest UUID from the config endpoint."""
@@ -11,9 +11,11 @@ async def _get_guest_uuid(client: httpx.AsyncClient) -> str:
         if config_response.status_code == 200:
             config_data = config_response.json()
             return config_data.get("guestUuid", "")
-    except Exception:
+    except (httpx.HTTPError, httpx.TimeoutException):
+        # Config endpoint may be unavailable; fallback to empty string
         pass
     return ""
+
 
 async def _check(email: str) -> Result:
     url = "https://lovescape.com/api/front/auth/signup"
@@ -36,7 +38,7 @@ async def _check(email: str) -> Result:
         async with httpx.AsyncClient(timeout=15.0) as client:
             # Step 1: Get guest UUID from config endpoint
             guest_uuid = await _get_guest_uuid(client)
-            
+
             # If no guest UUID, try to get it from the signup page
             if not guest_uuid:
                 try:
@@ -46,7 +48,8 @@ async def _check(email: str) -> Result:
                         uuid_match = re.search(r'guestUuid=([^;]+)', cookies)
                         if uuid_match:
                             guest_uuid = uuid_match.group(1)
-                except Exception:
+                except (httpx.HTTPError, httpx.TimeoutException):
+                    # Signup page may be unavailable; continue with empty UUID
                     pass
 
             # Step 2: Build payload with guest UUID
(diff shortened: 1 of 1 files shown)


Azaucifer:
Hi maintainers! 👋

Just a gentle ping on this PR. All checks have passed and the fix has been tested successfully.

**Quick recap:**
- The Lovescape API now requires a `guestUuid` field in signup requests
- This PR adds `_get_guest_uuid()` to fetch it from `/api/front/config`
- Includes `guestUuid` in the signup payload
- All checks: ✅ Passing

Let me know if there's anything else needed to get this merged!

Closes #613

kaifcodec:
@Azaucifer Really thanks for opening the PR, I was caught up with some important works so I was not able to review it.

Don't worry I will review it as soon as possible and let you know if it needs any changes.

kaifcodec:
In the meantime if you want to contribute further you are always welcome, you can add new modules or fix bugs in any existing modules!
I will be glad to review all of them when I get the time.

Azaucifer:
@kaifcodec Thanks for the response! No rush at all - I understand you're busy. Really enjoying working on this project!

I'll definitely look into contributing more modules or fixing other bugs. Are there any specific modules or areas that need attention right now?

Looking forward to the review whenever you have time! 🙌

kaifcodec:
@Azaucifer 
Currently, I don't think there are any obvious bugs to look for in the core. However, modules breaking because a target website changes its response structure is an ongoing issue with OSINT tools in general.

You can test both the email and username scans using the `--all` flag and see which sites are returning errors such as `403` or other unexpected responses. That can be a good indication that something has changed in their request headers, cookies, or API flow.

You can then inspect the affected site and capture the current request flow to determine whether it's caused by something like CAPTCHA/bot protection or if it's something we can fix. If it's fixable, you can just open a PR.

Azaucifer:
@kaifcodec That's a great suggestion! I'll run some scans with --all and look for modules returning errors.

I'll investigate any issues I find and open PRs for the fixable ones. Thanks for the guidance!

Will keep you posted on what I find. 🚀

kaifcodec (approved):
@Azaucifer Just reviewed it and this is perfect!

**MERGIING IT!**

Azaucifer:
@kaifcodec Thanks for the review and merge! 🎉

Really appreciate you taking the time to review it. Happy to contribute!

Looking forward to finding more issues to fix. 🚀

PR #624, opened 2026-09-02
fix(stackb): tolerate reordered and attributed markup

kristoisberg:
- Relax StackB login and profile HTML parsing in email and username modules

A recent (possibly today's) change broke the username and email modules.


kaifcodec (approved):
Merging it.

PR #626, opened 2026-09-02
feat: add Leanpub email and username modules

kristoisberg:
Username and email modules for the self-publishing site Leanpub. Uses the book creation endpoint, which also registers an account, but is not protected by a captcha, unlike the dedicated registration endpoint. An account is never actually created as the provided password does not match the requirements.


kaifcodec (approved):
LGTM.

PR #627, opened 2026-09-02
fix: improve Gumroad email and username detection

kristoisberg:
This PR makes several improvements to Gumroad modules:
- The email module now uses the registration form with an insufficiently long password, which never created an account. Previously, the module used the password reset form, but the module was not marked as loud. Now it is truly non-loud.
- The username module now parses detailed enrichments and has tighter restrictions on provided usernames.

### DIFF
--- a/user_scanner/email_scan/creator/gumroad.py
+++ b/user_scanner/email_scan/creator/gumroad.py
@@ -1,83 +1,64 @@
-import httpx
 import re
+
+from user_scanner.core.impersonate import impersonate_request_async
 from user_scanner.core.result import Result
 
-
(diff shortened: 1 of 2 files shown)


kaifcodec:
~@kristoisberg I tested the username module but it wasn't extracting any data for~ ~usernames it correctly returns `Found`, but no extra data scraped, I tested the~
~usernames johndoe, gumroad, ripper.~

~Check if the regex validation is correct or did they change their site structure.~

kaifcodec (approved):
@kristoisberg It is correct, I just wrongly testing on PR 626 branch.

PR #629, opened 2026-09-02
feat(user_scan): add Kickstarter creator module

kristoisberg:
Username module for Kickstarter.


kaifcodec:
@kristoisberg , during I noticed,

```
  [!] Kickstarter [https://www.kickstarter.com/profile/johndoe] (johndoe): Error (Kickstarter
profile markers were missing)
```
In case of real usernames but for non-existent it works fine, can you recheck it? Maybe response payload changes due to regional differences?

kristoisberg:
@kaifcodec should be sorted now, private profiles were not detected correctly

kaifcodec (approved):
MERGING IT.

PR #633, opened 2026-09-05
feat: add Rumble new username module, rename current module to rumble_channel

kristoisberg:
Rumble has two namespaces for usernames: users and channels, and channels don't necessarily need to be owned by an user with the same name. For example, https://rumble.com/c/ChadPrather is owned by an user different than https://rumble.com/user/chadprather (can be confirmed by comparing `creator_id` fields in module outputs). This PR adds support for both, renaming the current rumble module to rumble_channel and adding a new rumble module for users. It also adds additional enrichments to both modules.

### DIFF
--- a/user_scanner/user_scan/creator/rumble.py
+++ b/user_scanner/user_scan/creator/rumble.py
@@ -20,7 +20,10 @@ IMAGE_RE = re.compile(
     r'class="channel-header--(img|backsplash-img)"[^>]+src="([^"]+)"',
     re.IGNORECASE,
 )
-NOT_FOUND_MARKER = "<title>404 Not found</title>"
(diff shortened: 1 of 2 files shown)


kaifcodec:
@kristoisberg I found some edge cases:
1. For username scan (`.../u/`) when i scan `goodly` it goes well but in some usernames example johndoe it returns 410 Unexpected response body, maybe public or private account issue maybe.

2. In case of the (`.../c/`) similar thing happens, i.e. when channel exists it returns that 410 and when not it correctly returns `Not Found`.

kristoisberg:
@kaifcodec should be fixed now

kaifcodec (approved):
@kristoisberg Yeah it is fixed now. Thank you.

PR #636, opened 2026-09-09
fix(stackb): fix modules, add additional enrichments

kristoisberg:
- Fixed the username and email modules, which broke again (the site is being constantly updated right now, it seems)
- Migrated the username module to directly check profiles, rather than use the login form - previously, the site handled profile URLs case-sensitively, but they've changed that. Saves one request per existing account.
- Added additional enrichments to the username module based on recently added profile fields.
- General tidying.


kaifcodec (approved):
Perfect refactorisation @kristoisberg .

PR #637, opened 2026-09-09
feat: add Payhip email and username modules

kristoisberg:
This PR adds a username module and a loud email module for the e-commerce platform Payhip.

### DIFF
--- a/user_scanner/core/helpers.py
+++ b/user_scanner/core/helpers.py
@@ -56,6 +56,7 @@ LOUD_MODULES: Dict[str, List[str]] = {
         "cambly",
         "superlive",
         "medium",
+        "payhip",
     ],
 }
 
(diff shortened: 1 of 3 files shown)


kaifcodec:
@kristoisberg I just tested both the email and username modules and I only got 403 errors.
Can you check the issue?

kristoisberg:
> @kristoisberg I just tested both the email and username modules and I only got 403 errors. Can you check the issue?

I am only able to access it without a VPN myself... The site seems to be VERY aggressive in blocking VPNs and "test this website" type services: https://check-host.net/check-http?host=https%3A%2F%2Fpayhip.com%2F

On https://geopeeker.com, some locations aren't blocked. I believe residential proxies would also work, but I don't have access to any.

kaifcodec (approved):
@kristoisberg Ok, I am merging it.

PR #641, opened 2026-09-11
fix(core): group multi-module scans by category and prevent header bleeding

kaifcodec:
## Summary

When scanning multiple modules across different categories (e.g. via `-m mod1,mod2,...` or cross-scan), `_run_batch` previously executed all tasks in a single flat `asyncio.as_completed` pool without grouping. As network tasks completed out of order, once a category was marked in `printed_cats`, subsequent results from that category would not print a header and instead visually bleed under whatever unrelated category header completed most recently.

This PR fixes the issue in both `orchestrator.py` (`user_scan`) and `email_orchestrator.py` (`email_scan`):
- Bucket modules into category groups upfront while preserving order.
- Pre-spawn all tasks concurrently on the event loop to maintain full global parallel concurrency and semaphore limits.
- Await and stream results category-by-category so sites belonging to a category only print under their respective header.
- Add unit tests verifying category header grouping in multi-module runs.


json-hunter07 (approved):
I tested it looks fine @kaifcodec

PR #644, opened 2026-09-12
feat: add Wisio email and username modules

kristoisberg:
This PR adds username and email modules for [Wisio](https://www.wisio.com/).

### DIFF
--- a/user_scanner/email_scan/creator/wisio.py
+++ b/user_scanner/email_scan/creator/wisio.py
@@ -1,0 +1,24 @@
+import httpx
+
+from user_scanner.core.result import Result
+
+URL = "https://be-lb-app.scaleabout.com/api/v1/auth/email/is_registered"
+
+
+async def validate_wisio(email: str) -> Result:
+    show_url = "https://www.wisio.com"
+    try:
+        async with httpx.AsyncClient(timeout=15.0) as client:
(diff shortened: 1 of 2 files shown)


kaifcodec:
@kristoisberg I tested the email module which was perfect but in case of the username module, for usernames, `johndoe`, `alex` it is returning status code 404 Unexpected error (i.e. status didn't match) maybe the issue is private profile or something.

Anyways in case of other real usernames the user module works perfect.

kristoisberg:
> @kristoisberg I tested the email module which was perfect but in case of the username module, for usernames, `johndoe`, `alex` it is returning status code 404 Unexpected error (i.e. status didn't match) maybe the issue is private profile or something.
> 
> Anyways in case of other real usernames the user module works perfect.

This is a case where it's not possible to draw a definite conclusion...

The first request (is_username_allowed) reports "false" on both creator and follower accounts (I tested this out by creating one), and by the endpoint name, I assume that there can also be other allow/blocklist semantics in place.

The second request (user profile) only finds creator accounts, but does not find follower accounts.

I think the most reasonable thing to do is to return an error explicitly stating the ambiguous state, similarly to the TikTok module.

kaifcodec (approved):
That makes sense, merging it.

PR #645, opened 2026-09-12
feat(user_scan): add FredMiranda username module

kristoisberg:
Username module for the photography marketplace/forum [FredMiranda](https://www.fredmiranda.com/).

### DIFF
--- a/user_scanner/user_scan/creative/fredmiranda.py
+++ b/user_scanner/user_scan/creative/fredmiranda.py
@@ -1,0 +1,155 @@
+import html
+import re
+from urllib.parse import urlencode, urljoin
+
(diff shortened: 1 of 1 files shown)


kaifcodec:
@kristoisberg It is returning 403 in every case, test it again and let me know if the issue is only one my end.

kristoisberg:
> @kristoisberg It is returning 403 in every case, test it again and let me know if the issue is only one my end.

Most likely a region block again: https://check-host.net/check-http?host=https%3A%2F%2Fwww.fredmiranda.com%2F&csrf_token=1795ca2178993af1b1704e302342b890ca1833cb

kaifcodec (approved):
@kristoisberg Thanks for the review, MERGING IT now.

PR #649, opened 2026-09-14
fix: prevent duplicate CSV header on append

### LINKED ISSUE #648: CSV output (`-o file.csv -f csv`) duplicates the header row on every append run
### Summary
When using `-o file.csv -f csv` across multiple runs against the same output file, the header row gets written again on every append instead of just once. This corrupts the CSV, since the header ends up as a data row in the middle of the file.

### Where
- `user_scanner/core/formatter.py`, `into_csv()` (lines 18-19): always returns `CSV_HEADER + "\n" + rows`, with no way to get rows only.
- `user_scanner/__main__.py`, CSV branch (around lines 698-706): opens the output file in append mode and checks `has_content` only to decide whether to insert a blank line separator, never to decide whether the header should be skipped.

```python
# formatter.py
def into_csv(results):
    return CSV_HEADER + "\n" + "\n".join(result.to_csv() for result in results)
```

```python
# __main__.py
content_csv = formatter.into_csv(t_results)   # header baked in every time
...
with open(t_output, "a", encoding="utf-8") as f:
    if has_content:
        f.write("\n")
    f.write(content_csv)   # rewrites header on every append
```

### Reproduction
Ran the actual `Result`, `into_csv()`, and append logic twice against the same file:

username,category,site_name,status,url,extra,media,reason
alice,Social,TestSite,Found,,,,
username,category,site_name,status,url,extra,media,reason
alice,Social,TestSite,Found,,,,

### Suggested fix
Split `into_csv()` so it can return rows without the header, for example an optional `include_header` flag or a separate row joining function, and have `__main__.py` only write `CSV_HEADER` when `has_content` is `False`.

Sidu9:
## Description

Fixes an issue where the CSV header was duplicated every time results were appended to an existing output file.

Previously, the CSV formatter always included the header, causing repeated headers when the CLI was run multiple times against the same output file.

### Changes

- Ensure the CSV header is written only when the output file is new or empty.
- Keep CSV formatting focused on generating result rows.
- Preserve the existing append behavior for subsequent runs.

### Testing

- Verified that the header is written on the first run.
- Verified that subsequent runs append results without duplicating the header.
- Ran the existing test suite successfully.

Closes #648


kaifcodec:
Thanks for working on this @Sidu9!

The duplicate header issue when appending is definitely real. However, removing `CSV_HEADER` from `into_csv()` entirely breaks library usage for callers importing `into_csv()` from Python (as documented in `docs/USAGE.md`), who expect a valid CSV with headers.

Could we update this to:
1. Make `include_header: bool = True` an optional parameter in `into_csv(results, include_header=True)`.
2. In `__main__.py`, determine `has_content` first and call `formatter.into_csv(t_results, include_header=not has_content)`.
3. Add a test in `tests/test_formatter.py` verifying both `include_header=True` and `include_header=False`.

A-S-Manoj (approved):
LGTM. Tested locally, ran the append logic across multiple writes to the same file, and confirmed the header appears only once with no duplication. Output parses cleanly with `csv.DictReader`.

A regression test asserting the header shows up only once across repeated writes would be a nice add-on, but it's not blocking.

Approving.

A-S-Manoj:
@kaifcodec  Ah, missed your comment above, after checking, I agree with you. My approval was based on testing the CLI append path, which does work, but you're right that hardcoding into_csv() to drop the header breaks the documented library usage. +1 on the include_header param approach.

kaifcodec:
Yeah no worries @A-S-Manoj, it happens.
I know the core better so I caught that up. 
However thank you for opening the issue.

Sidu9:
Addressed your review @kaifcodec and added regression tests covering all the scenarios @A-S-Manoj. Let me know if this is Okay. Do we also need to make any changes to the docs to incorporate this new param?

A-S-Manoj:
Nice work @Sidu9, this addresses the concern.
Could you also update USAGE.md to mention the new `include_header` param?

Sidu9:
I've updated the documentation @A-S-Manoj. Let me know if its ok

chrisbrowns201392-create (approved):

kaifcodec (approved):
@Sidu9 Looks perfect now. MERGING IT. Thank you for the PR, you are welcome to contribute further to the project as well, like adding new site modules.

@A-S-Manoj Thank you again, for the detailed review on this.

PR #655, opened 2026-09-16
feat: add regional email provider domain enumeration

kristoisberg:
## Summary

Adds username-to-email enumeration for email OSINT scans.

Users can now pass `--email-domains` with username input to generate email candidates across provider scopes, then run the existing email scan modules against those generated addresses.

Example:

```bash
user-scanner -u johndoe --email-domains global
user-scanner -u johndoe --email-domains global,canada
user-scanner -u johndoe --email-domains usa
user-scanner -u johndoe --email-domains all
```

## What changed

- Added `--email-domains SCOPES` for `-u` and `-uf`.
- Added `--list-email-domains`.
- Added named provider scopes, including `global`, `usa`, `canada`, and many other country/local scopes.
- Added `all` as a shortcut for every configured scope.
- Reused the existing email scan flow, so `-m`, `-c`, exports, loud-module handling, and `--cross-scan` behave like normal email scans.
- Validates generated email addresses before scanning.
- Documents the new flags in `README.md` and `docs/FLAGS.md`.

## Notes

`--email-domains` is only valid with username input. It errors with `-e` or `-ef`, since email targets already include a domain.

Generated targets keep the existing multi-target export behavior, so exports split per generated email when an output path is provided.

## Verification

- `pytest -q`
- `python -m py_compile user_scanner/core/helpers.py user_scanner/__main__.py`
- `python -m user_scanner --list-email-domains`

## Follow-up

Provider lists are static. Add more domains by extending `EMAIL_DOMAIN_SCOPES`.


kaifcodec (approved):
@kristoisberg This is really a great feature. 
MERGING IT!

